PR #20, opened 2024-08-06
feat: add tile rendering job bundle example using Maya, Arnold, and ffmpeg

alichiba:
### Description

This job template is designed to submit a simple one frame tile rendering job using Maya, Arnold, and ffmpeg. This job bundle would be used with a modified Maya Arnold render handler, as specified in this [fork of the deadline-cloud-for-maya repository](https://github.com/alichiba/deadline-cloud-for-maya). It uses a bash script to call ffmpeg which requires Linux workers and access to the ffmpeg command from the workers. 

This template uses job parameter values (NumXTiles and NumYTiles) to specify the number of tiles which is used to define task parameter values (TileNumberX and TileNumberY) for the tile numbers, and is added to the runData for the Maya render step. A second step is defined to have a dependency on the render step, which uses a bash script to call ffmpeg to assemble the tiles. As stated in the comment in the bash script, the script only works with up to one layer of subdirectories and also assumes that the tiles in each folder are only for one frame, and are in the correct sequential order. 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/job_bundles/tile_render_maya_ffmpeg/template.yaml
+++ b/job_bundles/tile_render_maya_ffmpeg/template.yaml
@@ -263,7 +263,7 @@ steps:
           mode: NOTIFY_THEN_TERMINATE
 - name: tile assembly
   dependencies:
-  - dependsOn: masterLayer tile render
+  - dependsOn: tile render
   script:
     actions:
       onRun:
--- a/job_bundles/tile_render_maya_ffmpeg/template.yaml
+++ b/job_bundles/tile_render_maya_ffmpeg/template.yaml
@@ -1,0 +1,329 @@
+specificationVersion: jobtemplate-2023-09
+name: tile render maya ffmpeg
+parameterDefinitions:
+- name: MayaSceneFile
+  type: PATH
+  objectType: FILE
+  dataFlow: IN
+  userInterface:
+    control: CHOOSE_INPUT_FILE
+    label: Maya Scene File
+    groupLabel: Maya Settings
+    fileFilters:
+    - label: Maya Scene Files
+      patterns:
+      - '*.mb'
+      - '*.ma'
+    - label: All Files
+      patterns:
+      - '*'
+  description: The Maya scene file to render.
+- name: Frames
+  type: STRING
+  userInterface:
+    control: LINE_EDIT
+    label: Frames
+    groupLabel: Maya Settings
+  description: The frames to render. E.g. 1-3,8,11-15
+  minLength: 1
+- name: ProjectPath
+  type: PATH
+  objectType: DIRECTORY
+  dataFlow: NONE
+  userInterface:
+    control: CHOOSE_DIRECTORY
+    label: Project Path
+    groupLabel: Maya Settings
+  description: The Maya project path.
+- name: OutputFilePath
+  type: PATH
+  objectType: DIRECTORY
+  dataFlow: OUT
+  userInterface:
+    control: CHOOSE_DIRECTORY
+    label: Output File Path
+    groupLabel: Maya Settings
+  description: The render output path.
+- name: RenderSetupIncludeLights
+  type: STRING
+  userInterface:
+    control: CHECK_BOX
+    label: Include All Lights
+    groupLabel: Maya Settings
+  description: Include all lights in the render.
+  default: 'true'
+  allowedValues:
+  - 'true'
+  - 'false'
+- name: StrictErrorChecking
+  type: STRING
+  userInterface:
+    control: CHECK_BOX
+    label: Strict Error Checking
+    groupLabel: Maya Settings
+  description: Fail when errors occur.
+  default: 'false'
+  allowedValues:
+  - 'true'
+  - 'false'
+- name: OutputFilePrefix
+  type: STRING
+  userInterface:
+    control: LINE_EDIT
+    label: Output File Prefix
+    groupLabel: Maya Settings
+  description: The output filename prefix.
+- name: ImageWidth
+  type: INT
+  userInterface:
+    control: SPIN_BOX
+    label: Image Width
+    groupLabel: Maya Settings
+  minValue: 1
+  description: The image width of the output.
+- name: ImageHeight
+  type: INT
+  userInterface:
+    control: SPIN_BOX
+    label: Image Height
+    groupLabel: Maya Settings
+  minValue: 1
+  description: The image height of the output.
+- name: NumXTiles
+  type: INT
+  userInterface:
+    control: SPIN_BOX
+    label: Tiles across width
+    groupLabel: Tile Rendering Settings
+  minValue: 1
+  description: The number of tiles across the x axis.
+- name: NumYTiles
+  type: INT
+  userInterface:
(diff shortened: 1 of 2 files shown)


job_bundles/tile_render_maya_ffmpeg/template.yaml
```diff
@@ -0,0 +1,331 @@
+specificationVersion: jobtemplate-2023-09
+name: tile render maya ffmpeg
```
mwiebe: This is a great example, thanks for putting it together!

The PR summary covers a bit about what the job is for, but it would be great to have that right here in the template. A `description: ...` would be awesome to have here, that describes what the tiled job does, and why you may want to run a job as tiled.

Can you also adjust the name to match the pattern, like a document title, that the other examples follow? Something similar to "Maya Tiled Render"

job_bundles/tile_render_maya_ffmpeg/template.yaml
```diff
@@ -0,0 +1,331 @@
...
+  - name: Maya
+    description: Runs Maya in the background.
+    script:
+      embeddedFiles:
+      - name: initData
+        filename: init-data.yaml
+        type: TEXT
+        data: |
+          renderer: arnold
+          render_layer: masterLayer
```
mwiebe: This render layer would be good to have as a job parameter too.

job_bundles/tile_render_maya_ffmpeg/template.yaml
```diff
@@ -0,0 +1,331 @@
...
+    - name: runData
+      filename: run-data.yaml
+      type: TEXT
+      data: |
+        frame: {{Task.Param.Frame}}
+        camera: '{{Task.Param.Camera}}'
+        numXTiles: {{Param.NumXTiles}}
+        numYTiles: {{Param.NumYTiles}}
+        tileNumX: {{Task.Param.TileNumberX}}
+        tileNumY: {{Task.Param.TileNumberY}}
```
mwiebe, line 250: Rendering a selected tile isn't a typical parameter that renderers provide directly in their interface, but region rendering is. Because the purpose of the adaptor is to reflect the renderer interface, it would be better to modify the adaptor to directly accept the region{Min/Max}{X/Y} values. Adding region render support to the adaptor also feels like something we should be able to merge into the adaptor and release independently of this tile render job, so you could remove everything about building the custom branch.

Then you can put the transformation from tile coordinates into region bounds into the job bundle.

job_bundles/tile_render_maya_ffmpeg/template.yaml
```diff
@@ -0,0 +1,331 @@
+specificationVersion: jobtemplate-2023-09
+name: tile render maya ffmpeg
+parameterDefinitions:
+- name: MayaSceneFile
```
mwiebe, line 29: To make it easier to read and understand the parameters, some of our samples use a blank line to separate the differently grouped parameters, and start each block with a comment for the group label. See the parameters in https://github.com/aws-deadline/deadline-cloud-samples/blob/mainline/job_bundles/blender_render/template.yaml for an example.

job_bundles/tile_render_maya_ffmpeg/template.yaml
```diff
@@ -0,0 +1,331 @@
...
+          output_file_path: '{{Param.OutputFilePath}}'
+          render_setup_include_lights: {{Param.RenderSetupIncludeLights}}
+          strict_error_checking: {{Param.StrictErrorChecking}}
+          output_file_prefix: '{{Param.OutputFilePrefix}}'
+          image_width: {{Param.ImageWidth}}
+          image_height: {{Param.ImageHeight}}
+          error_on_arnold_license_fail: {{Param.ArnoldErrorOnLicenseFailure}}
+      actions:
+        onEnter:
+          command: MayaAdaptor
```
mwiebe: Could you switch this to the command `maya-openjd`?  We added the replacement name, but haven't done a full sweep to finish that.

job_bundles/tile_render_maya_ffmpeg/template.yaml
```diff
@@ -0,0 +1,331 @@
...
+        #!/bin/env bash
+
+        set -xeuo pipefail
+
+        # ASSUMPTION: output folder will either contain tiles, or folders with tiles, never both. Does not account for nested subdirectories.
+
+        has_subdirs=true
+
+        # If there are sub-directories in the output path, iterate over each item in the output file path directory and call ffmpeg for the sub-directory
+        for item in "{{Param.OutputFilePath}}"/*; do
```
mwiebe, line 294: Is this logic the reason you described it as "single frame" in the PR? The overall structure of this job looks great to handle a tiled render of an animation sequence, and it would be awesome to fix this part up to work with that too.

job_bundles/tile_render_maya_ffmpeg/template.yaml
alichiba: In regards to: 
_"Then you can put the transformation from tile coordinates into region bounds into the job bundle."_
How might you do that? I originally wanted to do this but struggled to come up with a way to do the calculations required for each tile's region coordinates such that each task would have access to the correct values based on the tile numbers.

Would another step need to run before the render step to calculate all the border values for each tile and store them in some sort of configuration file similar to how this is done with draft in deadline 10? If yes, how would this file be stored, and read when the task needs to specify the region coordinates?

alichiba: Another consideration; the changes I made to the adaptor includes creating the file name for the tiles, specific to the ffmpeg assembly pattern (top-down, left-right). If the logic moves to the job bundle, the filename would also need to be created there and passed into the adaptor.

mwiebe: Rather than try to explain it all in words, I wrote a sample git commit that's how I would tackle these things. Note, I have not run or tested this code in any way. But I hope it's helpful!

https://github.com/mwiebe/deadline-cloud-samples/commit/ca4fa1162f504c92798536e1931b6b1c6df90285

mwiebe: Thanks for adding the "next steps" in the description. Let's go ahead and merge this, and we'll be able to do a follow-up later that adjusts this.

job_bundles/tile_render_maya_ffmpeg/template.yaml
```diff
@@ -0,0 +1,391 @@
...
+        - run
+        - --connection-file
+        - '{{ Session.WorkingDirectory }}/connection.json'
+        - --run-data
+        - file://{{ Task.File.runData }}
+        cancelation:
+          mode: NOTIFY_THEN_TERMINATE
+- name: tile assembly
+  dependencies:
+  - dependsOn: masterLayer tile render
```
mwiebe: Rename this to `tile render` to match that step name now.

mwiebe (changes requested):

mwiebe (approved):

ddneilson (approved):

PR #28, opened 2024-09-05
fix: include bzip2 packages when reindexing

jflatt-gia:
Fixes: Reindex Conda channel job will not include bzip2 packages

Packages in the tar.bz2 format will not be included in the rebuild index

Include *.tar.bz2 in the wildcard glob

Full catalog of conda packages will be indexed

1. repodata.json includes tar.bz2 packages
2. reindex job eliminates tar.bz2 packages
3. repodata.json reinstated and includes tar.bz2 packages
4. this change applied
5. reindex job properly indexes tar.bz2 packages

Not applicable


mwiebe (approved):
Thanks for the fix!

lucaseck (approved):
Thank you for the additional fix @jflatt-gia!

PR #34, opened 2024-09-12
fix: update after effect job bundle

eherozhao:
After effects output param accepts the final output file path but current job passed in the output dir.

1. Added output file name param.
2. Updated aerender cmd
3. Updated README

Customers should be able to submit an AE job by this job template
Submit this job to Windows Service Managed Fleet with a custom conda channel with AE conda package and get the right output.

Updated the README.md

---


job_bundles/afterfx_render_one_task/template.yaml
```diff
@@ -73,15 +82,13 @@ steps:
...
+        command: bash
+        args: ['{{Task.File.Run}}']
+    embeddedFiles:
+    - name: Run
+      type: TEXT
+      data: |
+        mkdir -p '{{Param.OutputDirectory}}'
+
+        aerender -project '{{Param.ProjectFile}}' -comp '{{Param.CompName}}' \
+        -output '{{Param.OutputDirectory}}\{{Param.OutputFile}}' -s {{Param.StartFrame}} -e {{Param.EndFrame}}
```
mwiebe: I find commands like this more readable if they're split a bit more systematically by the option. Here's one way you could do that:

```
        aerender -project '{{Param.ProjectFile}}' \
            -comp '{{Param.CompName}}' \
            -output '{{Param.OutputDirectory}}\{{Param.OutputFile}}' \
            -s {{Param.StartFrame}} \
            -e {{Param.EndFrame}}
```

eherozhao: Thank you for the suggestion. Updated the template.

mwiebe (dismissed):

job_bundles/afterfx_render_one_task/template.yaml
```diff
@@ -73,15 +82,16 @@ steps:
...
-        - "{{Param.StartFrame}}"
-        - "-e"
-        - "{{Param.EndFrame}}"
+        command: bash
+        args: ['{{Task.File.Run}}']
+    embeddedFiles:
+    - name: Run
+      type: TEXT
+      data: |
+        mkdir -p '{{Param.OutputDirectory}}'
```
epmog: Since we're now running bash scripts, you'll want to add `set -euo pipefail` to the top of this bash script to ensure that an `aerender` failure bubbles up to the bash script failure, and thus the task will be failed rather than complete

job_bundles/afterfx_render_one_task/template.yaml
```diff
@@ -73,15 +82,16 @@ steps:
...
-        - "-output"
-        - "{{Param.OutputDirectory}}"
-        - "-s"
-        - "{{Param.StartFrame}}"
-        - "-e"
-        - "{{Param.EndFrame}}"
+        command: bash
+        args: ['{{Task.File.Run}}']
+    embeddedFiles:
+    - name: Run
```
epmog: nit: I'd recommend adding a file extension to the name that gets created. Makes it easier to inspect for debugging purposes. ie. your file editor will recognize the syntax for the file extension

PR #35, opened 2024-09-13
chore: use set instead of export in windows batch

jflatt-gia:
Fixes: Build complains about export command

Test phase under win-64 platform uses the batch file and complains about export command

Use set instead of export in windows batch

Environment variables are set

1. Build package under win-64, observe test phase complains about export, build succeeds anyways
2. Change export to set
3. Build package again, output is cleaner

Not applicable

- For instance, if applicable, has the sample's description been updated?

---


crowecawcaw:
We have a couple different patterns for conda build scripts in our samples and packages. In some cases for Windows, we've made `.bat` files that are meant to be run as Windows batch files. In some cases like this one, we've made a `.sh` file for Windows that's intended to be run with Git Bash for Windows. Using bash for both Linux and Windows scripts makes it a bit easier to support platforms at the same time.

So TLDR: I think these commands are correct and should be run with Git Bash.

mwiebe (approved):

mwiebe:
Reviewing this older PR, it looks like it is correct and we should apply it. The change is modifying the .bat files produced. Apologies for the confusion and delays in getting this in!

eherozhao (approved):

PR #36, opened 2024-09-24
fix: ripgrep changed executable location

jflatt-gia:
Fixes: ripgrep > v.11 moves .exe location, reintroduces WinError 206 (The filename or extension is too long)

My dependency tree now pulls in ripgrep 14 which has the rg.exe in a different location, breaking existing build

This just adds another location for removing rg.exe

Original location was left in, in case of existing builds

1. Previously successful recipe now gives FileNotFoundError: [WinError 206] The filename or extension is too long
2. Add this new location in the removal of rg.exe
3. Conda builds package successfully

Not applicable

- For instance, if applicable, has the sample's description been updated?

---


epmog (approved):
Thanks for putting up a fix!

Seems like there was some hope that ripgrep 14.0 would fix this with some long path support, but might be due to command line length limitations rather than long path shenanigans

moorec-aws (approved):

PR #40, opened 2024-10-28
feat: Add Cinema4D OpenJD adaptor recipe.

karthikbekalp:

This PR aims to help customers build their own Cinema 4D OpenJD Conda packages. 

Followed the examples of an existing recipe for Maya for Cinema4D. https://github.com/aws-deadline/deadline-cloud-samples/tree/mainline/conda_recipes/maya-openjd

Currently Deadline Cloud does not provide customers with Conda packages for Cinema 4D. By providing the Conda recipes, customers can build their own packages and use it. This can simplify the setup for our customers. 

To test this, I submitted the Conda recipe and built it in Deadline Cloud following a similar approach as described in https://aws.amazon.com/blogs/media/create-a-conda-package-and-channel-for-aws-deadline-cloud/ and I was able to get the renders for Cinema4D. 

I tested this locally using a Windows service managed fleet setup on DeadlineCloud and I was able to get the renders from Cinema4D. 
Screenshot 2024-10-28 at 10 15 47 AM
Screenshot 2024-10-28 at 10 14 22 AM

No, there is an upcoming PR for adding Conda recipes for Cinema 4D installer. I'll update the documentation for all of it then. 

---

### DIFF
--- a/conda_recipes/cinema4d-openjd/recipe/recipe.yaml
+++ b/conda_recipes/cinema4d-openjd/recipe/recipe.yaml
@@ -1,0 +1,55 @@
+# Recipe in rattler-build format.
+# See https://prefix-dev.github.io/rattler-build/latest/
+# Before building this recipe, build the conda 
+# packages for "deadline" and "openjd-adaptor-runtime" dependencies. 
+# Find these recipes for building the packages at : 
+# https://github.com/aws-deadline/deadline-cloud-samples/tree/mainline/conda_recipes
+
+context:
+  version: "0.4.1"
+
+package:
+  name: cinema4d-openjd
+  version: ${{ version }}
+
+# This source reference uses the source archive published to PyPI
+source:
+  url: https://pypi.io/packages/source/d/deadline-cloud-for-cinema-4d/deadline_cloud_for_cinema_4d-${{ version }}.tar.gz
+  sha256: 0404af9f6740a9778ced6f698c096ef52f6e8a0a73f975c4520727578a3dfabb
+  patches:
+    - 0001-Remove-version-hook.patch
+
+build:
+  number: 0
+  script:
+    - python -m pip install . -vv --no-deps --no-build-isolation
+
+requirements:
+  host:
+    - python
+    - hatchling
+    - hatch-vcs
+    - pip
+  run:
+    - python
+    - deadline 0.48.*
+    - openjd-adaptor-runtime 0.8.*
+
+tests:
+- python:
+    imports:
+      - deadline.cinema4d_adaptor
+- requirements:
+    run:
+      - pip
+  script:
+    - pip check
+    - cinema4d-openjd --help
+
+about:
+  summary: AWS Deadline Cloud for Cinema 4D
+  homepage: https://github.com/aws-deadline/deadline-cloud-for-cinema-4d
+  license: Apache-2.0
+  license_file:
+    - LICENSE
+    - NOTICE
--- a/conda_recipes/cinema4d-openjd/deadline-cloud.yaml
+++ b/conda_recipes/cinema4d-openjd/deadline-cloud.yaml
@@ -1,0 +1,32 @@
+condaPlatforms:
+  - platform: linux-64
+    variant: py311
+    defaultSubmit: true
+    variantConfig:
+      python:
+      - 3.11
+    buildTool: rattler-build
+  - platform: linux-64
+    variant: py312
+    defaultSubmit: true
+    variantConfig:
+      python:
+      - 3.12
+    buildTool: rattler-build
+  - platform: win-64
+    variant: py311
+    defaultSubmit: true
+    variantConfig:
+      python:
+      - 3.11
+    buildTool: conda-build
+  - platform: win-64
+    variant: py312
+    defaultSubmit: true
+    variantConfig:
+      python:
+      - 3.12
+    buildTool: conda-build
+jobParameters:
+- name: CondaChannels
+  value: conda-forge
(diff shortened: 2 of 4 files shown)


conda_recipes/cinema4d-openjd/recipe/deactivate.sh
```diff
@@ -0,0 +1,8 @@
+unset DEADLINE_CLOUD_PYTHONPATH
```
crowecawcaw: Outside the scope of this PR, but we might need to choose different env var names here in the future. KeyShot's adapter uses the same one. While it's not a common use case, if both adapters were in the environment at the same thing, they'd stop on each others env vars.

crowecawcaw (dismissed):

conda_recipes/cinema4d-openjd/recipe/deactivate.sh
karthikbekalp: Nice catch, this environment variable will be removed soon. I'm waiting [on this PR ](https://github.com/aws-deadline/deadline-cloud-for-cinema-4d/pull/73) to get merged after which we should `C4DPYTHONPATH311` instead.

conda_recipes/cinema4d-openjd/recipe/activate.bat
```diff
@@ -0,0 +1,13 @@
...
+REM We need to add additional paths for pywin32 dependency.
+set C4DPYTHONPATH311="%%CONDA_PREFIX%%\Lib\site-packages;%%CONDA_PREFIX%%\Lib\site-packages\win32;%%CONDA_PREFIX%%\Lib\site-packages\win32\lib;%%CONDA_PREFIX%%\Lib\site-packages\Pythonwin"
+
+REM You can add your license environment variables for Cinema 4D and Redshift here. 
+REM For licensing Cinema 4D, you can use the g_licenseServerUrl variable. 
+REM https://support.maxon.net/hc/en-us/articles/1500006333301-R21-License-Server-for-R21-and-onward-automatic-user-login
+REM set g_licenseServerUrl=http://[ip address]:[port]
+REM For licensing Redshift, you can use the redshift_LICENSE environment variable. 
+REM https://help.maxon.net/c4d/en-us/#../Subsystems/Default/Content/html/RLM+Floating+Licenses.html#RLMFloatingLicenses-Option1-EnvironmentVariables
+REM set redshift_LICENSE=[port]@[ip-address]
```
mwiebe: I don't think this block comment should be here. As a general rule, mixing configuration like license servers and ports with software packages is not a good structure to use. Something like queue environments (https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/configure-jobs.html) is better suited to the task.

conda_recipes/cinema4d-openjd/recipe/activate.sh
```diff
@@ -0,0 +1,13 @@
+export DEADLINE_CLOUD_PYTHONPATH="$CONDA_PREFIX\\Lib\\site-packages"
+export DEADLINE_CINEMA4D_EXE="$CONDA_PREFIX\\c4d\\Commandline.exe"
+
+# We need to add additional paths for pywin32 dependency.
+export C4DPYTHONPATH311="$CONDA_PREFIX\\Lib\\site-packages;$CONDA_PREFIX\\Lib\\site-packages\\win32;$CONDA_PREFIX\\Lib\\site-packages\\Pythonwin;$CONDA_PREFIX\\Lib\\site-packages\\win32\\Lib"
```
mwiebe: Maybe this could be incorporated into the Cinema4D conda build recipe instead of the adaptor?

conda_recipes/cinema4d-openjd/recipe/activate.sh
```diff
@@ -0,0 +1,13 @@
+export DEADLINE_CLOUD_PYTHONPATH="$CONDA_PREFIX\\Lib\\site-packages"
+export DEADLINE_CINEMA4D_EXE="$CONDA_PREFIX\\c4d\\Commandline.exe"
```
mwiebe: This is a property that should be set by the Cinema4D package, not here. That package could choose a different prefix for installation, for example, and this adaptor wouldn't work with it anymore.

mwiebe (changes requested):

conda_recipes/cinema4d-openjd/recipe/activate.bat
karthikbekalp: Makes sense, it was simpler to use the license environment variables in my testing but queue environments definitely looks better suited. I'll update this.

conda_recipes/cinema4d-openjd/recipe/activate.sh
karthikbekalp: I agree 100%. I'll update this.

conda_recipes/cinema4d-openjd/recipe/activate.sh
karthikbekalp: Hmm, the pywin32 dependencies are added by the `openjd-adaptor-runtime` which is installed during the installation of this adaptor. So should we still try to build this in the Cinema 4D conda build recipe?

joel-wong-aws (dismissed):

mwiebe (dismissed):

conda_recipes/cinema4d-openjd/recipe/recipe.yaml
```diff
@@ -0,0 +1,51 @@
...
+requirements:
+  host:
+    - python
+    - hatchling
+    - hatch-vcs
+    - pip
+  run:
+    - python
+    - deadline 0.48.*
+    - openjd-adaptor-runtime 0.8.*
```
crowecawcaw, line 36: Do these packages need to be in the conda channel for this to work? Or do they get pulled from pip automatically? Or maybe conda has a copy of all PyPi libraries? I don't think we've published the Deadline library to conda forge.

karthikbekalp: > Do these packages need to be in the conda channel for this to work?

Customers would have to also build the `deadline` and `openjd-adaptor-runtime` packages before building this package. These recipes are in the https://github.com/aws-deadline/deadline-cloud-samples/tree/mainline/conda_recipes. Maybe it makes more sense to add a readme so that its clearer?

crowecawcaw (approved):

joel-wong-aws (approved):

PR #42, opened 2024-11-09
feat: Add support for rattler-build in the sample conda tools

mwiebe:

The rattler-build tool is much faster at building conda packages than conda-build. See https://prefix.dev/blog/rattler_build_on_conda_forge to read about how this is available on conda-forge now.

* Currently Linux-only, add the ability to use rattler-build to build conda packages submitted with the submit-conda-build script. Because rattler-build doesn't support S3 conda channels directly, it creates proxy channels using mountpoint-s3.
* Define a way to specify the build tool as either conda-build or rattler-build, and update the README.md documentation to describe that.
* Add glob platform matching to the submit-package-build script to make it easier to submit variants based on patterns.
* Update the sample conda package recipes to include a recipe.yaml for rattler-build. The deadline-cloud.yaml metadata specifies rattler-build for Linux and conda-build for Windows.
* Update the Blender 4.2 sample to 4.2.3.
* Fix a STDOUT vs STDERR bug in the package build script, and refactor its logging to speed up debugging issues like this in the future.

Experimentation shows roughly 4 or 5 times faster package builds on some of the samples, so switching will save a lot of time.

Submitted all the sample package recipes to a build farm with various options, and confirmed successful builds.

Updated the README.md documentation about how to submit package builds.

---

### DIFF
--- a/conda_recipes/README.md
+++ b/conda_recipes/README.md
@@ -2,15 +2,19 @@
 
 ## Summary
 
-This directory contains sample tools for creating an S3 conda channel and building new packages into it
-on AWS Deadline Cloud.
+This directory contains sample tools for creating an S3 conda channel and
+building new packages for either Linux or Windows into it on AWS Deadline Cloud.
 
 * The job bundle [conda_build_linux_package](conda_build_linux_package) defines a job that
-  can build packages for Linux.
-* The command `submit-package-job` submits a job for building a provided conda build recipe on
-  a specified set of conda platforms. It takes the job bundle for Linux, and extends it for the set of
-  conda platforms.
-* Conda build recipe samples to submit with `submit-package-job`.
+  is cross-platform but configured for Linux.
+* The submission command `submit-package-job` submits a job for running
+  a provided conda build recipe on a specified set of conda platforms. It takes the job
+  bundle, and edits it to match the arguments provided.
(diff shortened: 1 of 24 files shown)


conda_recipes/README.md
```diff
@@ -131,6 +148,15 @@ You can add more files, like a LICENSE.txt to document the license of the recipe
...
 
+#### The buildTool option
+
+You can select the default build tool between conda-build and rattler-build for the whole recipe
+by setting this option.
+
+```
+buildTool: conda-build
+```
+
```
crowecawcaw: It might help to include a bit more info about rattler-build in one of these docs sections. For example:
- rattler-build is generally faster than conda-build
- rattler-build has a slightly different recipe format so recipes are not 1-1 compatible (though they are very similar)
- any limitations to rattler-build besides being linux only?

crowecawcaw (dismissed):

conda_recipes/README.md
mwiebe: I've added some more info here, thanks for the review!

crowecawcaw (approved):

leongdl (approved):
I will try this with Wine builds today.

rena-cheng (approved):

PR #44, opened 2024-11-12
fix: Improve robustness of conda queue envs and package build

mwiebe:

In exploratory testing, we've encountered some corner cases that were not handled in the package build scripts and queue environments.

* When removing an environment directory, confirm that it was removed before proceeding, as permissions or a process lock could prevent that from succeeding.
* During environment activation, turn off bash strict variable checking, as some activation scripts expect `$VAR` to resolve to "" when it is undefined.
* When checking the time since last env update, check for the timestamp file first and assume 0 minutes if it is missing.

Anyone using these samples will have fewer errors, and better messages when errors do occur.

Updated a test farm's queue environment, then submitted all the sample package build recipes to it.

No, it has no change to functionality.

---

### DIFF
--- a/conda_recipes/conda_build_linux_package/scripts/enter-package-build-env.sh
+++ b/conda_recipes/conda_build_linux_package/scripts/enter-package-build-env.sh
@@ -29,7 +29,13 @@ function conda_clean_on_error {
     if [ ! "$1" = "0" ]; then
         echo "Error detected, removing the $ENV_NAME environment and cleaning the Conda cache."
         for ENVS_DIR in $(conda info --json | python -c "import json, sys; v = json.load(sys.stdin); print('\n'.join(v['envs_dirs']))"); do
-            rm -rf $ENVS_DIR/$ENV_NAME
+            if [ -d $ENVS_DIR/$ENV_NAME ]; then
+                echo "Removing directory $ENVS_DIR/$ENV_NAME for the environment"
+                rm -rf $ENVS_DIR/$ENV_NAME
+                if [ -d $ENVS_DIR/$ENV_NAME ]; then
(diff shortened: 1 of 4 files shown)


conda_recipes/conda_build_linux_package/scripts/enter-package-build-env.sh
```diff
@@ -29,7 +29,13 @@ function conda_clean_on_error {
     if [ ! "$1" = "0" ]; then
         echo "Error detected, removing the $ENV_NAME environment and cleaning the Conda cache."
         for ENVS_DIR in $(conda info --json | python -c "import json, sys; v = json.load(sys.stdin); print('\n'.join(v['envs_dirs']))"); do
-            rm -rf $ENVS_DIR/$ENV_NAME
+            if [ -d $ENVS_DIR/$ENV_NAME ]; then
+                echo "Removing directory $ENVS_DIR/$ENV_NAME for the environment"
+                rm -rf $ENVS_DIR/$ENV_NAME
+                if [ -d $ENVS_DIR/$ENV_NAME ]; then
+                    echo "WARNING: Could not remove the directory. Possibly a permissions error or a process holding a lock."
+                fi
```
jairaws, line 37: Q: Are we missing an `exit 1` here, or is this intended?

jairaws (approved):

conda_recipes/conda_build_linux_package/scripts/enter-package-build-env.sh
mwiebe: This is intended, so that the rest of the cleanup code runs as well.

ddneilson (approved):

PR #45, opened 2024-11-12
docs: Add readme on how to build V-Ray conda package

AlexTranAmz:
We want to have clear instruction on how customers could use V-Ray conda recipe we provide to build conda package
Add a README.md to lay out some instruction specifically for V-Ray
We have better instruction to better guide customers to successfully build V-Ray conda package
N/A since this is only a README.md update

Yes

---

### DIFF
--- a/conda_recipes/vray/README.md
+++ b/conda_recipes/vray/README.md
@@ -1,0 +1,17 @@
+# V-Ray conda build recipe
+
+## Decide what version of the archive file to download
+- Download `vraystd_adv_62022_rhel8_clang-gcc-11.2` for **x86**
+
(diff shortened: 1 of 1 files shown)


conda_recipes/vray/README.md
```diff
@@ -0,0 +1,17 @@
+# Maya 2024 conda build recipe
```
mwiebe: nit: Need to update this title.

mwiebe (changes requested):

conda_recipes/vray/README.md
AlexTranAmz: Changed!

mwiebe (approved):

epmog (approved):

PR #47, opened 2024-11-14
feat: Add a sample conda package recipe for Maya and MtoA

mwiebe:

While bringing applications to Deadline Cloud service-managed fleets is open-ended with custom [queue environments](https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/provide-applications.html#provide-applications-other-package), its default configuration consumes conda packages. The best way to bring your own applications distributed with installers for use by conda is not obvious, but a set of sample package recipes provides a great starting point, and we want to grow the set of samples.

Implement a sample package recipe for Autodesk Maya 2025 that works both on Linux and Windows. Add a second sample for MtoA, the Arnold renderer plugin for Maya, that integrates with the first one.

* Linux recipes are using rattler-build, Windows recipes using conda-build
* Supports both Linux and Windows
* The Linux recipe uses the upstream artifact from Autodesk
* The Windows recipe includes instructions on how to create a source artifact in the recipe README
* Both recipes include many comments to explain their structure
* Modify the Maya/Arnold turntable sample to use Maya >= 2024 instead of requiring specifically 2024.

Filed an issue with rattler-build about its behavior excluding .pyc files: https://github.com/prefix-dev/rattler-build/issues/1191

Customers can create their own Maya package, and can use the example 

Submitted package builds, and then submitted the Maya/Arnold turntable sample to render with those packages. Verified the newly built packages were used, and that the output video is as expected.

The recipe includes README content.

---

### DIFF
--- a/conda_recipes/maya-2025/README.md
+++ b/conda_recipes/maya-2025/README.md
@@ -1,0 +1,76 @@
+# Maya 2025 conda build recipe
+
+## Instructions for Maya plugin packages
+
+This Maya conda build recipe configures the `MAYA_MODULE_PATH` environment variable
+to include several paths to search for plugin `.mod` files. When creating a plugin
+package, place its `.mod` file in one of these so that Maya loads the plugin at startup.
+
+* `$PREFIX/usr/autodesk/maya$MAYA_VERSION/modules`
+* `$PREFIX/usr/autodesk/modules/maya/$MAYA_VERSION`
+* `$PREFIX/usr/autodesk/modules/maya`
+
+## Download the archive file for Linux
+
+Download the Autodesk_Maya_2025_Linux_64bit.tgz full download file from Autodesk, and
+place it in the `conda_recipes/archive_files` directory in your git clone of the
+[https://github.com/aws-deadline/deadline-cloud-samples](deadline-cloud-samples) repository for
+submitting package build jobs,
+
+## Creating an archive file for Windows
+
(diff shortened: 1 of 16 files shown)


karthikbekalp (approved):

conda_recipes/maya-2025/README.md
```diff
@@ -0,0 +1,76 @@
...
+for use by the conda build recipe.
+
+1. Launch a fresh Windows Server 2022 instance.
+    1. From the AWS EC2 management console, select the option to Launch instance.
+    2. Enter instance name "Create Windows Maya archive".
+    3. Select "Microsoft Windows Server 2022 Base" for the AMI. (As of writing, Windows Server 2025 did not work with these instructions.)
+    4. Select an instance type with enough vCPUs and RAM, for example c5.4xlarge has 8 vCPUs and 16 GiB RAM.
+    5. Select "Proceed without a key pair" for the "Key pair (login)" option.
+    6. We will use SSM port forwarding to avoid sending RDP protocol traffic directly over the internet.
+        1. Make sure that "Allow RDP traffic" is unchecked.
```
npmacl, line 34: why do we not allow RDP traffic here if we connect with RDP in step 2?

mwiebe: Because we're documenting how to set it up with SSM port forwarding that sends the traffic through an encrypted tunnel instead of directly over the internet.

npmacl (approved):

PR #50, opened 2024-11-19
feat: Add V-Ray job bundle template

AlexTranAmz:
We have provided conda recipe to build V-Ray conda package. Having sample job bundle for V-Ray help customer better utilize V-Ray
Add job bundle for V-Ray
Customer could use their newly build V-Ray conda package to render

Submit multiple different `.vrscene` with this `template.yaml` using `deadline bundle gui-submit`

Yes

---

### DIFF
--- a/job_bundles/vray_render/README.md
+++ b/job_bundles/vray_render/README.md
@@ -1,0 +1,13 @@
+# V-Ray sample job bundle
+Use this job bundle to create a V-Ray job to use with AWS Deadline Cloud. \
+Read more about job bundle and how to use it with AWS Deadline Cloud [here](../README.md). \
+Read more about how to customize this job bundle bellow
+
+## Prerequisite
+- Have V-Ray conda package host on a conda channel. \
+Read more about how to create V-Ray conda package [here](../../conda_recipes/vray/README.md)
+- Have a sample `.vrscene` and its dependencies files.
+
+## Job bundle customization
+- All available customization flags could be found on [Chaos guide](https://docs.chaos.com/display/VNS/V-Ray+Standalone+Command+Line+Options) to tailor this job bundle
+- Sample command line for V-Ray could be found on [Chaos example](https://docs.chaos.com/display/VNS/Getting+Started+with+Commands)
--- a/job_bundles/vray_render/template.yaml
+++ b/job_bundles/vray_render/template.yaml
@@ -30,23 +30,22 @@ parameterDefinitions:
     groupLabel: Render Parameters
   default: "./output"
   description: Choose the render output directory.
-- name: OutputPattern
+- name: OutputFilename
   type: STRING
   userInterface:
     control: LINE_EDIT
-    label: Output File Pattern
+    label: Output File Name
     groupLabel: Render Parameters
-  default: "output_####"
-  description: Enter the output filename pattern (without extension).
+  default: "output"
+  description: Enter the output filename (without extension).
 - name: Format
   type: STRING
   userInterface:
-    control: DROPDOWN_LIST
+    control: LINE_EDIT
     label: Output File Format
     groupLabel: Render Parameters
-  description: Choose the file format to render as.
-  default: JPEG
-  allowedValues: [TGA, RAWTGA, JPEG, IRIS, IRIZ, PNG, HDR, TIFF, OPEN_EXR, OPEN_EXR_MULTILAYER, CINEON, DPX, DDS, JP2, WEBP]
+  default: png
+  description: Enter the file format to render as.
 
 # Software Environment
 - name: CondaPackages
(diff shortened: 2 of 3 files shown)


YutongLi291 (dismissed):

job_bundles/vray_render/template.yaml
```diff
@@ -0,0 +1,85 @@
...
+  description: Enter the output filename pattern (without extension).
+- name: Format
+  type: STRING
+  userInterface:
+    control: DROPDOWN_LIST
+    label: Output File Format
+    groupLabel: Render Parameters
+  description: Choose the file format to render as.
+  default: JPEG
+  allowedValues: [TGA, RAWTGA, JPEG, IRIS, IRIZ, PNG, HDR, TIFF, OPEN_EXR, OPEN_EXR_MULTILAYER, CINEON, DPX, DDS, JP2, WEBP]
```
mwiebe: This doesn't look like a list of file extensions, e.g. the extension for OpenEXR is ".exr". Does V-Ray document the list that should be here?

job_bundles/vray_render/template.yaml
```diff
@@ -0,0 +1,85 @@
...
+    groupLabel: Render Parameters
+  default: "./output"
+  description: Choose the render output directory.
+- name: OutputPattern
+  type: STRING
+  userInterface:
+    control: LINE_EDIT
+    label: Output File Pattern
+    groupLabel: Render Parameters
+  default: "output_####"
```
mwiebe: What's the purpose of `####` here?

mwiebe (changes requested):

job_bundles/vray_render/template.yaml
AlexTranAmz: Thanks for catching this. Fixed in the new commit

job_bundles/vray_render/template.yaml
AlexTranAmz: The idea is to have a pattern if need to render multiple frames, but V-Ray have a flag for number frames or not -> do not need the pattern anymore, just need the output file name

job_bundles/vray_render/template.yaml
```diff
@@ -0,0 +1,84 @@
...
+    groupLabel: Render Parameters
+  default: "./output"
+  description: Choose the render output directory.
+- name: OutputFilename
+  type: STRING
+  userInterface:
+    control: LINE_EDIT
+    label: Output File Name
+    groupLabel: Render Parameters
+  default: "output"
```
mwiebe: With the current approach, it's probably more natural to include the extension here as `"output.png"` and remove the `Format` parameter.

AlexTranAmz: Got it! Will fix in the next commit

mwiebe (approved):

job_bundles/vray_render/README.md
```diff
@@ -0,0 +1,13 @@
+# V-Ray sample job bundle
+Use this job bundle to create a V-Ray job to use with AWS Deadline Cloud. \
+Read more about job bundle and how to use it with AWS Deadline Cloud [here](../README.md). \
+Read more about how to customize this job bundle bellow
```
sakshie95, line 4: nit: typo ``bellow``

job_bundles/vray_render/template.yaml
```diff
@@ -0,0 +1,76 @@
...
+  default: "./output"
+  description: Choose the render output directory.
+- name: OutputFilename
+  type: STRING
+  userInterface:
+    control: LINE_EDIT
+    label: Output File Name
+    groupLabel: Render Parameters
+  default: "output.png"
+  description: Enter the output filename (with extension of choice).
```
sakshie95, line 40: Do we support all output file extensions? Or do we need a list here

job_bundles/vray_render/template.yaml
```diff
@@ -0,0 +1,76 @@
...
+          # Configure the task to fail if any individual command fails.
+          set -xeuo pipefail
+          vray -sceneFile='{{Param.VraySceneFile}}' \
+               -imgFile='{{Param.OutputDir}}/{{Param.OutputFilename}}' \
+               -display=0
+  hostRequirements:
+    attributes:
+    - name: attr.worker.os.family
+      anyOf:
+      - linux
```
sakshie95, line 76: I see this is a linux specific job bundle, do we need that as a parameter of the template name in case we want to add for more os in the future?

sakshie95 (changes requested):

YutongLi291 (approved):

job_bundles/vray_render/template.yaml
AlexTranAmz: I could add that in the next PR. We plan to have another PR right after this to add more V-Ray field in and we can add OS into it as well.

job_bundles/vray_render/template.yaml
AlexTranAmz: we do not know the list since V-Ray standalone do not have a list. Talked with Mark and we decided to make default the generic file extension and let customer decide based on their need.

job_bundles/vray_render/README.md
AlexTranAmz: Will fix in the next PR along with adding more fields like mention below

sakshie95 (approved):

PR #56, opened 2024-11-26
feat: Add fleet health check CloudFormation template

ishasat:
We want to provide customers with a sample solution for continuous health check monitoring for a single Deadline Cloud fleet.

Publish a sample CloudFormation template with supporting documentation for how to set up continuous health check monitoring.

Customers can use this template directly to setup health check monitoring for their Deadline Cloud fleets.

I deployed this template to my account where I have a customer-managed fleet set up. In my initial CMF setup, workers were not being created successfully which caused the worker health check alarm to be triggered. After I fixed my CMF configuration, I saw workers spin up successfully and the alarm recovered.

Yes, I added a README.

---

### DIFF
--- a/cloudformation/farm_templates/cmf_templates/README.md
+++ b/cloudformation/farm_templates/cmf_templates/README.md
@@ -1,0 +1,65 @@
+# Deploying Deadline Cloud fleet health check
+
+## Introduction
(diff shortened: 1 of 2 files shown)


eng-josh (approved):

eherozhao (approved):
Are we going to have a high level README for the whole CloudFormation folder? I think we need it for indexing purpose.

PR #57, opened 2024-12-03
feat: add after effects conda recipe

eherozhao:
We don't have AE conda recipe in the sample repo.

Provided AE 25.0 Conda Recipe as a guideline for customers to use

Customers can use the recipe to build the Conda Package.

Submitted the recipe to the package build queue to build the package and then submit an AE job to the production queue to consume the package.


conda_recipes/aftereffects-25.0/README.md
```diff
@@ -0,0 +1,46 @@
+# Adobe After Effects 25 conda build recipe
+
+## Creating an archive file for Windows
+
+The Windows installer requires Administrator permissions that are not available in most conda package
+build environments, such as on a Deadline Cloud service-managed fleets. Follow these instructions to
+install Adobe After Effects 25 on a freshly created EC2 instance as Administrator, and create an archive file
+for use by the conda build recipe.
```
lucaseck: Should a note be added that this is one way to create the archive? This could very easily also be created on any Windows machine they may have physically no?

conda_recipes/aftereffects-25.0/recipe/meta.yaml
```diff
@@ -0,0 +1,27 @@
+{% set name = "aftereffects" %}
+{% set version_partial = "25.0" %}
+{% set version_minor = "1" %}
```
lucaseck: Is the 0 in 25.0 not the minor version and 1 is the patch version?

eherozhao: hmm you are right. Will update.

conda_recipes/aftereffects-25.0/README.md
eherozhao: Will update.

lucaseck (approved):

conda_recipes/aftereffects-25.0/README.md
```diff
@@ -0,0 +1,45 @@
...
+   3. Select "Microsoft Windows Server 2022 Base" for the AMI.
+   4. Select an instance type with enough vCPUs and RAM, for example c5.4xlarge has 8 vCPUs and 16 GiB RAM.
+   5. Select "Proceed without a key pair" for the "Key pair (login)" option.
+   6. Make sure that "Allow RDP traffic" is unchecked. We will use SSM port forwarding to avoid sending RDP
+      protocol traffic directly over the internet.
+   7. Set the storage to at least 64 GiB. Adjust other settings as you like, e.g. if you want an encrypted volume of type gp3.
+   8. Select "Launch instance."
+   9. If it asks, select "Proceed without key pair" and proceed with the launch.
+   10. Once it launched, navigate to the instance detail page. Select "Connect," and with "Session manager" selected, again select "Connect."
+       If it says "SSM Agent is not online," you may have to wait a few minutes for it to initialize.
```
viknith, line 21: Someone else was hitting an issue with this where the SSM agent was not activating despite waiting a very long time. Please sync with them in case these instructions need to be updated with troubleshooting notes.

conda_recipes/aftereffects-25.0/recipe/build_win.sh
```diff
@@ -0,0 +1,45 @@
...
+cat <<EOF > "$PREFIX/etc/conda/activate.d/$PKG_NAME-$PKG_VERSION-vars.bat"
+set "AE_LOCATION=$PREFIX/opt/aftereffects"
+set "AE_VERSION=$AE_VERSION"
+set "PATH=$PREFIX/opt/aftereffects;%PATH%"
+EOF
+cat "$PREFIX/etc/conda/activate.d/$PKG_NAME-$PKG_VERSION-vars.bat"
+
+cat <<EOF > $PREFIX/etc/conda/activate.d/$PKG_NAME-$PKG_VERSION-vars.sh
+export "AE_VERSION=$AE_VERSION"
+export "AE_LOCATION=\$CONDA_PREFIX/opt/aftereffects"
```
viknith, line 27: Silly question, why are the AE_LOCATION and PATH variables being set with different syntax in the `.sh` file compared to the `.bat` file? Aren't they the same values?

conda_recipes/aftereffects-25.0/recipe/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+{% set version_minor = "0" %}
+{% set version = version_partial + "." + version_minor %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+- url: archive_files/Adobe_AfterEffects_{{ version.replace(".", "_") }}_Windows_installation.zip # [win64]
+  sha256: f181bddce95e0f45555c5188a9a515a43ce20d0071d566286f748cbba36b5716 # [win64]
```
viknith, line 12: I'm assuming this would be replaced with the hash that customers generate from their own zip file?

viknith (approved):

conda_recipes/aftereffects-25.0/recipe/meta.yaml
eherozhao: The hash should be the same for the same version of the archive.

conda_recipes/aftereffects-25.0/README.md
eherozhao: I don't think that is the scope of this recipe. We can link a SSM official guide here but I honestly don't think this is very helpful.

conda_recipes/aftereffects-25.0/recipe/build_win.sh
eherozhao: because the syntax of bash is not the same as windows bat.

PR #60, opened 2024-12-06
chore: add needed package for Maya 2025 to work with V-Ray

AlexTranAmz:
When building V-Ray for Maya, we identify some missing dependencies for Maya.
In particular, Qt6 package from Maya need some dependencies such as:
- `libxcb-icccm`
- `libxcb-cursor`
- `libxcb-keysyms`
- `libxkbcommon-x1`
We need to add those dependencies in to ensure Maya functionality. In specific, we need:
- xcb-util-wm for `libxcb-icccm`
- xcb-util-cursor for `libxcb-cursor`
- xcb-util-keysyms for `libxcb-keysyms`
- libxkbcommon-x11 for `libxkbcommon-x1`
Maya 2025 will have enough dependencies to make it work reliably 

`./submit-package-job maya-2025`
Submit a Maya job + a V-Ray for Maya job and was able to render successfully
Yes

---

### DIFF
--- a/conda_recipes/maya-2025/recipe/build.sh
+++ b/conda_recipes/maya-2025/recipe/build.sh
@@ -27,7 +27,9 @@ ln -r -s "$INSTALL_DIR/bin/maya$MAYA_VERSION" "$INSTALL_DIR/bin/maya"
 # from the system package manager, dnf.
 mkdir -p "$SRC_DIR/download"
 cd "$SRC_DIR/download"
-dnf download --resolve -y freetype alsa-lib fontconfig harfbuzz libbrotli graphite2 libxkbfile
+dnf download --resolve -y freetype alsa-lib fontconfig harfbuzz libbrotli graphite2 \
+    libxkbfile xcb-util-cursor xcb-util-wm xcb-util-keysyms libxkbcommon-x11
+
 for RPM_FILE in *.rpm; do
     rpm2cpio "$RPM_FILE" | cpio -idm
 done


conda_recipes/maya-2025/recipe/build.sh
```diff
@@ -27,7 +27,9 @@ ln -r -s "$INSTALL_DIR/bin/maya$MAYA_VERSION" "$INSTALL_DIR/bin/maya"
 # from the system package manager, dnf.
 mkdir -p "$SRC_DIR/download"
 cd "$SRC_DIR/download"
-dnf download --resolve -y freetype alsa-lib fontconfig harfbuzz libbrotli graphite2 libxkbfile
+dnf download --resolve -y freetype alsa-lib fontconfig harfbuzz libbrotli graphite2 \
```
leongdl, line 30: Have we asked autodesk  why more and more dependencies are needed? You are adding X windows dependencies, which gets very heavy in the long run.

leongdl: If every DCC now starts to need these X binaries, it is much better to just put it in to the base AMI.

AlexTranAmz: - I believe this is need to in the case of V-Ray rather than Maya (Maya will work without this but V-Ray for Maya will not work).

- We could add this into V-Ray conda recipe and link Maya to it, but this pose some potential error: Modifier could easily change path inside conda recipe build script and it will break V-Ray for Maya.

- What we could do is document this and see if only V-Ray need this or lots of DCC need it and make a decision to add it to the base AMI or not.

leongdl: Yes - document is the key here so we can see what patterns of dependencies are needed in the SMF worker image vs conda recipie.

The more we put into the conda package, the more vulnerabilities we have to patch by hand

leongdl (approved):

conda_recipes/maya-2025/recipe/build.sh
mwiebe: Could we include the additional ones in the v-ray package? It's just those binaries with the dependency, right?

AlexTranAmz: We could, but as I mentioned above, we will then have to patchelf a Maya .so file to these binaries lib which will be an issue if customers wants to change the path from default.

mwiebe: The Maya .so files should be fine, we  just have to apply that to the vray ones. What were you finding that made it different?

AlexTranAmz: The thing that's make it different is when invoke Maya, it was able to run. But then Maya need some dependencies to be run to call V-Ray, which in this case it's not runable due to missing dependencies.

By adding these package to V-Ray conda recipe rather than here, we need to link a specific .so file from Maya (not Maya or Render .so file) to a lib we store those downloaded dependencies on V-Ray conda package, which seem unstable in my opinion. 

Also since all those packages only needed for Maya and have nothing to do with V-Ray (V-Ray executable do not need it), leaving it in V-Ray seem like a not idea choice

mwiebe: Thanks for updating the description with those additional details, that helps understand the change.

mwiebe (approved):

PR #70, opened 2025-01-07
docs: Clarify that the starter_farm S3 bucket must already exist

mwiebe:

If someone starts deploying the starter_farm bucket without stepping through the README, they
can miss that the S3 bucket should exist before they deploy the template.

* Updated the description and GUI metadata for the JobAttachmentsBucketName to be more clear that its for an existing bucket.

It should be more obvious from the label of the parameter and the description of it, that the bucket should exist already.

Confirmed that the adjusted label and description display when deploying the template from the AWS console.

N/A

---


baxeaz (approved):

sakshie95 (approved):

leongdl (commented):
LGTM

PR #79, opened 2025-02-01
feat: Conda recipe for RSMB 6.4.4 for After Effects, with README about plugin packaging

cilevitz:
After Effects requires a project's referenced plugins to be installed onto the render machine in a location that requires elevated permissions on Windows to modify. After Effects has a version-dependent plugin path in the system registry, which also requires elevated permissions to modify. We need to install plugin files to a location that After Effects can access at render time in a service-managed fleet environment, where elevated permissions are not available.

We have created a workflow for producing Conda packages that dynamically install plugins into an SMF job's render environment. After Effects can then load a submitted project's referenced plugins at render time.

Users can add supported plugins to their render jobs so their effects will render correctly in an SMF environment.

The workflow was tested on CPU-only and GPU-accelerated service-managed fleets with RSMB 6.4.4 as the packaged plugin. We confirmed that following the workflow gets farm resources set up correctly and submitted jobs produce correct render output on Adobe After Effects 24.0 and 25.1.

Yes, there is a README.md in the recipe folder describing the workflow and its prerequisites.

### Is this a breaking change?
No.

----

### DIFF
--- a/conda_recipes/rsmb-6.4.4/recipe/build_win.sh
+++ b/conda_recipes/rsmb-6.4.4/recipe/build_win.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+set -xeuo pipefail
+
+# The location within $PREFIX where rsmb will be installed
+mkdir -p $PREFIX/opt
+cp -r $SRC_DIR/rsmb $PREFIX/opt/
+
+
+# See https://docs.conda.io/projects/conda/en/latest/dev-guide/deep-dives/activation.html
+# for details on activation. The Deadline Cloud sample queue environments use bash
+# to activate environments on Windows, so we recommend always producing both .bat and .sh files.
+
+mkdir -p "$PREFIX/etc/conda/activate.d"
+mkdir -p "$PREFIX/etc/conda/deactivate.d"
+
+cat <<EOF > "$PREFIX/etc/conda/activate.d/$PKG_NAME-$PKG_VERSION-vars.bat"
+robocopy "$PREFIX/opt/rsmb" "$PREFIX/opt/aftereffects/Support Files/Plug-Ins" *.aex
+EOF
+cat "$PREFIX/etc/conda/activate.d/$PKG_NAME-$PKG_VERSION-vars.bat"
+
+cat <<EOF > "$PREFIX/etc/conda/deactivate.d/$PKG_NAME-$PKG_VERSION-vars.bat"
+del "$PREFIX\opt\aftereffects\Support Files\Plug-ins\RSMB_64.aex"
+del "$PREFIX\opt\aftereffects\Support Files\Plug-ins\RSMBPro_64.aex"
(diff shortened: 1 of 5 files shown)


cilevitz:
Moved from https://github.com/aws-deadline/deadline-cloud-for-after-effects/pull/95

Please see previous comment thread: https://github.com/aws-deadline/deadline-cloud-for-after-effects/pull/95#issuecomment-2591127287

Merged README.md into the rsmb-6-4-4 recipe folder.

viknith:
Does this plugin depends on After Effects being installed first? Can you verify and test whether you should add AE as some kind of dependency like here?

https://github.com/aws-deadline/deadline-cloud-samples/blob/mainline/conda_recipes/maya-mtoa-2025/recipe/meta.yaml#L25-L29

Relevant documentation for that is here: https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/create-conda-recipe-mtoa-plugin.html

conda_recipes/rsmb-6.4.4/recipe/build_win.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# The location within $PREFIX where rsmb will be installed
+mkdir -p $PREFIX/opt
+cp -r $SRC_DIR/rsmb $PREFIX/opt/
+
+
+# See https://docs.conda.io/projects/conda/en/latest/dev-guide/deep-dives/activation.html
+# for details on activation. The Deadline Cloud sample queue environments use bash
+# to activate environments on Windows, so we recommend always producing both .bat and .sh files.
+
```
viknith, line 12: Also the .sh activate/deactivate scripts are missing

eherozhao: https://github.com/aws-deadline/deadline-cloud-samples/blob/mainline/conda_recipes/aftereffects-25.0/recipe/build_win.sh#L25-L30 here is what you can use as an example.

conda_recipes/rsmb-6.4.4/recipe/build_win.sh
```diff
@@ -0,0 +1,26 @@
...
+mkdir -p "$PREFIX/etc/conda/activate.d"
+mkdir -p "$PREFIX/etc/conda/deactivate.d"
+
+cat <<EOF > "$PREFIX/etc/conda/activate.d/$PKG_NAME-$PKG_VERSION-vars.bat"
+robocopy "$PREFIX/opt/rsmb" "$PREFIX/opt/aftereffects/Support Files/Plug-Ins" *.aex
```
viknith, line 17: This folder should be `Plug-ins` not `Plug-Ins`

viknith: Screenshot 2025-02-17 at 9 51 21 PM

viknith:
We have an alternate PR that goes through how to do an AE plugin Conda recipe that's free and doesn't require licensing. Ideally we can reference that to make improvements to this PR: https://github.com/aws-deadline/deadline-cloud-samples/pull/88

PR #89, opened 2025-03-28
fix: fixed AE_LOCATION env variable to support BYO AE plugins on SMF

viknith:
The plugin recipe was made based on the Conda recipe for After Effects in the samples repository. However, we support AE on SMF, so we should match the AE location to what we have on SMF so that customers face minimal issues and can use the recipe out-of-the-box.

Fix the AE location

Customers can use the plugin conda recipe to add plugin support for AE SMF or CMF without any further modifications based on what we have in this repository

Submitted job with BYO Saber and SMF AE, jobs succeeded!

Yes

---


MarinMaksutaj (dismissed):

conda_recipes/aftereffects-saber/recipe/build_win.sh
```diff
@@ -1,9 +1,8 @@
 #!/bin/sh
 set -xeuo pipefail
 
-mkdir -p $PREFIX/opt
+AE_LOCATION="$PREFIX/aftereffects"
```
mwiebe: You should be able to remove this from the recipe, and instead use the `aftereffects` plugin as a dependency. The variable `AE_LOCATION` will then get defined when the host environment is activated.

mwiebe (approved):

jaikish (approved):

PR #91, opened 2025-03-28
chore: Update Cinema 4D open JD version recipe to 0.7.3

karthikbekalp:

The conda recipe for Cinema 4D open JD was outdated. 

Updated the hash and the version for the recipe. 

Customers can now build and use the new conda packages. 

Built the packages and ran renders on the built packages in Cinema 4D. 

image

No

---

### DIFF
--- a/conda_recipes/cinema4d-openjd/recipe/meta.yaml
+++ b/conda_recipes/cinema4d-openjd/recipe/meta.yaml
@@ -4,7 +4,7 @@
 # https://github.com/aws-deadline/deadline-cloud-samples/tree/mainline/conda_recipes
 
 {% set name = "cinema4d-openjd" %}
-{% set version = "0.7.2" %}
+{% set version = "0.7.3" %}
 
 package:
   name: {{ name }}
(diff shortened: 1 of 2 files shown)


conda_recipes/cinema4d-openjd/recipe/meta.yaml
```diff
@@ -40,7 +40,7 @@ requirements:
   run:
     # For the variant config file in deadline-cloud.yaml to work, python must not specify a version here.
     - python
-    - deadline ==0.49.*
+    - deadline >=0.49.7
```
mwiebe: This should likely be `>=0.49.7,<0.50` to keep constraining within backwards compatible versions.

karthikbekalp: Makes sense, updated.

mwiebe (dismissed):

mwiebe (approved):

joel-wong-aws (approved):

PR #109, opened 2025-07-15
docs: Host configuration reboot example script. 

leongdl:
- Deadline Cloud added support for Host Configuration scripts for Service Managed workers. [Feature](https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/worker-host.html)
- This feature runs scripts as "root", and allows actions such as rebooting a worker. This is especially useful for changing drivers, or other OS level changes.

- Add documentation

- Add documentation
- 
- Read the Documentation

- This PR is documentation

---


host_configuration_scripts/worker_reboot/linux.sh
```diff
@@ -0,0 +1,11 @@
+echo Shutting down
+ls /var/lib/deadline
+if [ -f /var/lib/deadline/rebooted ]; then
+    echo "Already Rebooted, ready to start"
+    exit 0
+fi
```
mwiebe, line 8: Does the host configuration script run a second time after the reboot?

host_configuration_scripts/worker_reboot/linux.sh
```diff
@@ -0,0 +1,11 @@
+echo Shutting down
```
mwiebe: nit: This will print even if it is not shutting down in the "already rebooted" case.

host_configuration_scripts/worker_reboot/linux.sh
leongdl: Yes - as long as we do not return 0 as part of the worker protocol design.

Normally, it transitions to the `STOPPED` end state and then the worker shutsdown. For the reboot case, it goes to `STOPPED` state and then restarts, so it will run again.

host_configuration_scripts/worker_reboot/linux.sh
leongdl: Oh good point! Let me fix this.

Sorry forgot about this PR, will fix it today.

host_configuration_scripts/worker_reboot/linux.sh
leongdl: I will also put up for review the Windows version.

host_configuration_scripts/worker_reboot/linux.sh
```diff
@@ -0,0 +1,11 @@
...
+ls /var/lib/deadline
+if [ -f /var/lib/deadline/rebooted ]; then
+    echo "Already Rebooted, ready to start"
+    exit 0
+fi
+touch /var/lib/deadline/rebooted
+chmod 660 /var/lib/deadline/rebooted
```
leongdl, line 14: This file is opened so jobs can also check and know the host has been setup. (or contain any data like driver version)

leongdl: I also added a comment to make it more explicit.

host_configuration_scripts/worker_reboot/linux.sh
leongdl: Updated!

host_configuration_scripts/worker_reboot/linux.sh
leongdl: Updated! I have moved the comment around. Added a short pre-amble of what the script does, and the print outs when it reboots and when it is ready to process jobs.

host_configuration_scripts/worker_reboot/README.md
```diff
@@ -0,0 +1,6 @@
+# This directory provides example scripts to reboot a Deadline Cloud Service Managed Worker
```
jairaws, line 1: Do we plan to add more worker rebooting scripts? I feel this category should be broader. Something like `worker_management`.

host_configuration_scripts/worker_reboot/linux.sh
```diff
@@ -0,0 +1,15 @@
...
+ls /var/lib/deadline
+if [ -f /var/lib/deadline/rebooted ]; then
+    echo "Host already Rebooted, ready to start."
+    exit 0
+fi
+echo "Host has not been rebooted, Rebooting worker host."
+touch /var/lib/deadline/rebooted
+# Optional chmod, use 660 if jobs need to check if this host is configured.
+chmod 660 /var/lib/deadline/rebooted
+sudo reboot now
```
jairaws, line 16: If something fails with the file creation, the worker will go into a doom spiral. Can we check that the file was created before rebooting? We should also log a message calling out that the file was created.

host_configuration_scripts/worker_reboot/README.md
```diff
@@ -0,0 +1,6 @@
...
+
+Worker reboots maybe required for system configuration changes. For example, installing new GPU device drivers, or joining network domains.
```
jericht:
```suggestion
Worker reboots may be required for system configuration changes. For example, installing new GPU device drivers, or joining network domains.
```

host_configuration_scripts/worker_reboot/linux.sh
jericht: @leongdl Is the Windows version up yet? Have you pushed it?

host_configuration_scripts/worker_reboot/README.md
leongdl: Not specifically - each use case example will be a specific folder to keep it simple.

host_configuration_scripts/worker_reboot/README.md
leongdl: Good call!

host_configuration_scripts/worker_reboot/linux.sh
leongdl: I will make a separate PR. I need to do more testing on it.

host_configuration_scripts/worker_reboot/linux.sh
leongdl: Sure, I'll add a message.

I don't think we need to check the file creation since this is a SMF specific feature and this will work. Otherwise the script is cumbersome. 

If customers edit this script, we also work with a shared responsibility model for customers to debug it.

jericht (dismissed):

host_configuration_scripts/worker_reboot/linux.sh
```diff
@@ -0,0 +1,18 @@
+echo "Running Host Configuration script to reboot this worker." \
+     "If already rebooted, start processing steps." \
+     "Otherwise, reboot"
+ls /var/lib/deadline
+if [ -f /var/lib/deadline/rebooted ]; then
+    echo "Host already Rebooted, ready to start."
+    exit 0
+fi
+echo "Host has not been rebooted."
+# Caution When changing this file or path.
```
jairaws, line 10: nit: typo `When` -> `when`

jairaws (approved):

jericht (approved):

PR #84, opened 2025-02-25
feat: added recipe for redshift-for-houdini on SMF

yuanmich2:
### What was the problem/requirement? (What/Why)

There was no example for Redshift on Houdini.

### What was the solution? (How)

This pr adds a conda recipe for Redshift on Houdini

### What is the impact of this change?

Now everyone can see an example of how to build a conda package for Redshift on Houdini

### How was this change tested?

I performed some tests on my own farm

### Was this change documented?

The sample comes with its own readme

---

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

PR #110, opened 2025-07-16
feat: add houdini redshift sample Conda recipe

lucaseck:
This is a continuation of #84 and thanks to @yuanmich2 for the starting point.
We have a Houdini sample now in the repository and Redshift is commonly used with Houdini, but had no sample of its own.
- Add a Redshift for Houdini conda recipe sample
- Also a one line change to the Houdini sample so that we have the full MAJOR.MINOR.PATCH version in the HOUDINI_VERSION env var that it sets. Originally this was so that I could match the redshift version exactly by reading that env var, but I've instead gone with calling `houdini --version` to get than info. I still think it's valuable to have the full version string in HOUDINI_VERSION however
- Edit the Houdini package README to better clarify how other plugin packages should register themselves with Houdini installed by that package.

Users can more easily self serve Redshift for Houdini via Conda

- Ran render tests from Houdini+Redshift using the package and verified the outputs as well as looked at the render logs to make sure everything looked to be functioning correctly
- Manually tested the version detection logic with a variety of different version strings to make sure the best choice was always selected
- Verified that the package worked seamlessly with the Houdini package built using the sample in this repository

- A README has been added

---

### DIFF
--- a/conda_recipes/houdini-redshift-2025/recipe/build.sh
+++ b/conda_recipes/houdini-redshift-2025/recipe/build.sh
@@ -1,0 +1,106 @@
+#!/bin/sh
+set -xeuo pipefail
+
+# The version without the update number
+REDSHIFT_VERSION="$PKG_VERSION"
+
+INSTALLER_DIR="$SRC_DIR/installer"
+INSTALLER="$INSTALLER_DIR/redshift_${REDSHIFT_VERSION}_linux_x64.run"
+REDSHIFT_UNPACK_DIR="$INSTALLER_DIR/redshift_installer_artifacts"
+REDSHIFT_ROOT="$PREFIX/opt/redshift"
+
+# Extract the run file into its accompanying tarball and setup script
+chmod u+x "$INSTALLER"
+$INSTALLER --target "$REDSHIFT_UNPACK_DIR" --noexec
+
+# Skip superuser check by modifying the setup script in-place
+sed -i 's/\[ "$(id -u)" != "0" \]/false/' "$REDSHIFT_UNPACK_DIR/setup.sh"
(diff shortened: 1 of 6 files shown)


conda_recipes/houdini-redshift-2025/recipe/build.sh
```diff
@@ -0,0 +1,115 @@
...
+REDSHIFT_VERSION=$PKG_VERSION
+
+INSTALLER_DIR="$SRC_DIR/installer"
+INSTALLER="$INSTALLER_DIR/redshift_${REDSHIFT_VERSION}_linux_x64.run"
+REDSHIFT_UNPACK_DIR="$INSTALLER_DIR/redshift_installer_artifacts"
+REDSHIFT_ROOT="$PREFIX/opt/redshift"
+REDSHIFT_PACKAGE_DIR="$REDSHIFT_ROOT/redshift_package"
+
+# Extract the run file into its accompanying tarball and setup script
+chmod u+x $INSTALLER
```
justinsaws: Nit: anytime you are expanding a variable you should quote it. It helps prevent any unintentional code injection.

justinsaws (dismissed):

conda_recipes/houdini-redshift-2025/recipe/build.sh
lucaseck: Was making some other changes so fixed this as well(and I think all other cases).

erico-aws (dismissed):

justinsaws (dismissed):

justinsaws (approved):

erico-aws (approved):

PR #131, opened 2025-09-10
docs: minor AE docs update for plugin documentation

viknith:
Our existing AE docs for conda recipes don't point to plugin docs and we want the folder naming clarity to be better so that viewers don't see AE Saber and get dissuaded by it.

Update docs

Better docs

No need

---


erico-aws:
Can you add a section to the plugin bundle sample's README that summarizes how it adds multiple plugins to AE?

erico-aws (approved):

karthikbekalp (approved):

PR #132, opened 2025-09-11
feat: add sample for listing conda packages

crowecawcaw:
It's sometimes useful to see what conda packages are available.

Add a quick job sample that prints the list of available packages from a channel.

Helps exploring the service-managed packages and debugging conda.

I submitted the job and looked at the package list.

Yes
---


lucaseck (approved):

lucaseck (changes requested):
LGTM, but I think usually new bundles are added as `feat:` not `chore:`

lucaseck (approved):

epmog (approved):

PR #134, opened 2025-09-15
feat: add fast build optimization and extra args for conda builds

kenyrish-amazon:
Conda/rattler builds were taking a long time.

Can add parameters when submitting conda/rattler builds to skip compression, speeding up build times. This speeds up builds by up to 40%.

Also, added an option to add extra arguments to the conda/rattler builds

When running `submit-package-job`, users can add the `--fast-build` flag to speed up package job builds.
Users can also add the `--extra-build-tool-args` option to add additional arguments to the conda/rattler builds.

Tested with and without `--fast-build` flag, and with conda and rattler builds. Made sure that the builds were faster with the flag, and everything worked as expected when the flag was not included.

Yes, added to the readme.

---

### DIFF
--- a/conda_recipes/README.md
+++ b/conda_recipes/README.md
@@ -131,7 +131,7 @@ $ ./submit-package-job blender-4.2 --s3-channel s3://<another-s3-bucket>/channel
 ### Enabling fast build optimizations
 
 You can enable fast build optimizations to create faster builds
-by adding the `--fast-build` flag:
+by adding the `--fast-build` or `-f` flag:
 
 ```
 $ ./submit-package-job blender-4.2 --fast-build
@@ -143,6 +143,20 @@ This enables:
 
 The fast build optimization is particularly beneficial for packages with many files or large binaries,
 as it reduces package size and can speed up both the build process and package installation.
+
+### Adding custom build arguments
+
+You can pass additional arguments to the conda-build or rattler-build command using the `--extra-build-tool-args` or `-a` option:
+
+```
+$ ./submit-package-job blender-4.2 --extra-build-tool-args "--no-test --quiet"
+```
+
+This allows you to customize the build process with any supported conda-build or rattler-build arguments:
+- **conda-build**: Examples include `--no-test`, `--quiet`, `--debug`
+- **rattler-build**: Examples include `--quiet`, `--debug`, `--skip-existing`
+
+The build arguments are parsed as space-separated values and added to the build command. Use quotes to group arguments that contain spaces.
 
 ## Recipe directory structure for `submit-package-build`
 
--- a/conda_recipes/README.md
+++ b/conda_recipes/README.md
@@ -138,7 +138,7 @@ $ ./submit-package-job blender-4.2 --fast-build
 ```
 
 This enables:
-- **conda-build**: Uses `--zstd-compression-level 1` for faster compression with good compression ratio
+- **conda-build**: Uses `--zstd-compression-level 1` for faster compression with a larger package size
 - **rattler-build**: Uses `--package-format conda:min` for optimized package format
 
 The fast build optimization is particularly beneficial for packages with many files or large binaries,
--- a/conda_recipes/README.md
+++ b/conda_recipes/README.md
@@ -128,6 +128,22 @@ IAM role of the queue you're submitting to includes permissions for the S3 bucke
 $ ./submit-package-job blender-4.2 --s3-channel s3://<another-s3-bucket>/channel/prefix
 ```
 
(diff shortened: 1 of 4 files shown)


conda_recipes/README.md
```diff
@@ -128,6 +128,22 @@ IAM role of the queue you're submitting to includes permissions for the S3 bucke
...
+
+You can enable fast build optimizations to create faster builds
+by adding the `--fast-build` flag:
+
+```
+$ ./submit-package-job blender-4.2 --fast-build
+```
+
+This enables:
+- **conda-build**: Uses `--zstd-compression-level 1` for faster compression with good compression ratio
```
leongdl: Nit: faster compression but still "good" ? 

How much larger was the file? I think it was like 3:1 ratio? What was it before?

conda_recipes/README.md
```diff
@@ -128,6 +128,22 @@ IAM role of the queue you're submitting to includes permissions for the S3 bucke
 $ ./submit-package-job blender-4.2 --s3-channel s3://<another-s3-bucket>/channel/prefix
 ```
 
+### Enabling fast build optimizations
+
+You can enable fast build optimizations to create faster builds
+by adding the `--fast-build` flag:
```
leongdl: Super nit, I'm trying to think of a good name for this.

--fast-build-low-compression (-fb)? 

Just tossing ideas around

conda_recipes/submit-package-job-script.py
```diff
@@ -414,6 +422,9 @@ def main():
     parser.add_argument(
         "--all-platforms", action="store_true", help="Submit all the platforms specified by the recipe's deadline-cloud.yaml."
     )
+    parser.add_argument(
+        "--fast-build", action="store_true", help="Enable build optimizations for faster package creation."
```
leongdl: Can we add more to the help text? EG: it uses zstd level 1, and what this actually does?

leongdl: Can we also add a quick option like `-fb` or `-f` ?see the `-p, -q` above.

conda_recipes/README.md
kenyrish-amazon: The files become 1.5-2x larger without the compression.

conda_recipes/conda_build_linux_package/scripts/build-package.py
```diff
@@ -276,8 +276,13 @@ def main():
             sys.exit(1)
         prefix_length_option = ["--prefix-length", f"{args.override_prefix_length}"]
 
+    # Check for fast build optimization environment variable
+    enable_fast_build = os.environ.get("CONDA_BUILD_ENABLE_FAST_BUILD", "false").lower() == "true"
```
mwiebe: The other options are all passed as CLI options, not environment variables. Could you change this to follow that pattern as well? That way someone reading the code can see the whole interface in one place. If we split the interface to use both CLI options and environment variables for different options, it will be harder to understand and debug issues.

conda_recipes/conda_build_linux_package/scripts/build-package.py
```diff
@@ -276,8 +276,13 @@ def main():
...
+
     # Run the package build tool
     if args.build_tool == "conda-build":
+        fast_build_opts = ["--zstd-compression-level", "1"] if enable_fast_build else []
```
mwiebe, line 291: What about adding an option like "ExtraBuildToolArgs" so that users could have more ways to customize things? E.g. they could use compression level 2, or they could use an entirely different argument.

conda_recipes/README.md
kenyrish-amazon: I wanted to keep it as fast-build, since any other optimizations can be added under this flag as well

conda_recipes/submit-package-job-script.py
kenyrish-amazon: Added `-f` option, and some more info to the help text. I didn't add all the zstd info so it wouldn't get too long, that is in the readme as well.

conda_recipes/conda_build_linux_package/scripts/build-package.py
kenyrish-amazon: Modified this, passed it as a cli as well.

conda_recipes/conda_build_linux_package/scripts/build-package.py
kenyrish-amazon: Added this option, users can call it with `--extra-build-tool-args`

benl-2023 (approved):

sakshie95 (approved):

PR #140, opened 2025-09-26
feat: Added support for Maya2026 to Redshift2025 conda recipe

sakshie95:
Redshift 2025 recipe does not support Maya2026

Added support for Maya2026 in the Redshift recipe. Changes include adding a few required paths to the rpaths using patchelf

Customers can use the redshift recipe with Maya2026

Built the conda packages for redshift 2025 and maya2026 using the starter farm on local and verified that a test render works successfully.

redshiftmayascene 0001

Built the conda packages for redshift 2025 and maya2025 too using the starter farm on local and verified that a test render works successfully for testing backwards compatibility :

redshiftmayascene 0001

- For instance, if applicable, has the sample's description been updated? N/A

---

### DIFF
--- a/conda_recipes/maya-redshift-2025/recipe/recipe.yaml
+++ b/conda_recipes/maya-redshift-2025/recipe/recipe.yaml
@@ -27,7 +27,7 @@ build:
 
 requirements:
   run:
-    - maya >=2025, <2026
+    - maya >=2025, <2027
 
 
 about:
--- a/conda_recipes/maya-redshift-2025/README.md
+++ b/conda_recipes/maya-redshift-2025/README.md
@@ -1,5 +1,12 @@
 # Redshift for Maya conda build recipe
 
+This package provides Redshift 2025.4.2 support for Maya versions 2025, and 2026.
+
+## Supported Maya Versions
+
+- Maya 2025  
+- Maya 2026
+
 ## Download the installer file for Linux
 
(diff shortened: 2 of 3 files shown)


conda_recipes/maya-redshift-2025/recipe/recipe.yaml
```diff
@@ -27,7 +27,7 @@ build:
 
 requirements:
   run:
-    - maya >=2025, <2026
```
kenyrish-amazon, line 30: Maybe add to the readme indicating the redshift-2025 recipe works with Maya 2026, since we dont have a maya-redshift-2026 folder?

benl-2023 (dismissed):

conda_recipes/maya-redshift-2025/recipe/recipe.yaml
sakshie95: I actually made the changes to the readme then removed it from the PR. Mostly cz I wasn't sure if we should be writing we support 2022-2026. I've personally tested 2024-2026 so I'm comfortable in writing that. Thoughts?

sakshie95: After a discussion we decided to update this to 2025-2026

AlexTranAmz (approved):

benl-2023 (approved):

PR #141, opened 2025-10-15
chore: Added links between the C4D base sample and its plugins.

erico-aws:

The Cinema4D base recipe doesn't mention plugins and should link to the plugin samples.

Added a brief section to the README that links to the plugin samples.

Better discoverability of the plugin samples.

N/A

No.

---

### DIFF
--- a/conda_recipes/cinema4d-2024/README.md
+++ b/conda_recipes/cinema4d-2024/README.md
@@ -49,3 +49,10 @@ for use by the conda build recipe.
        `Write-S3Object -BucketName MY_BUCKET_NAME -Key Cinema4D_2024_2024.5.1_Win.zip -File Cinema4D_2024_2024.5.1_Win.zip`.
 4. From the AWS EC2 management console, select the instance you used and terminate it.
 5. Download the zip file to the `conda_recipes/archive_files` directory in your git clone of the [deadline-cloud-samples](https://github.com/aws-deadline/deadline-cloud-samples) repository for submitting package build jobs, and update the Windows source artifact hash in the Cinema 4D-2024 conda build recipe meta.yaml.
+
+## Instructions for Cinema4D plugin packages
+This repository contains examples for how to create plugin conda packages for Cinema4D (DISCLAIMER: These samples were only tested with 2025):
+
+* [Cinema4D to Arnold](../cinema4d-c4dtoa-2025/)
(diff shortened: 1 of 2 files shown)


viknith:
There is already a PR for After Effects to do this that hasn't gotten enough reviewers to merge: https://github.com/aws-deadline/deadline-cloud-samples/pull/131

conda_recipes/cinema4d-2024/README.md
```diff
@@ -1,5 +1,12 @@
 # Cinema 2024 conda build recipe
 
+## Instructions for Cinema4D plugin packages
+Plugins can be placed into the Cinema4D's plugin directory. eg. `"$PREFIX/cinema4d/bin/plugins"` or `%PREFIX%\cinema4d\plugins`. This repository contains examples for how to do this with Cinema4D 2025:
```
karthikbekalp: We currently haven't tested our plugin conda recipes with 2024. Maybe we can add a disclaimer that we have only tested with 2025?

conda_recipes/cinema4d-2025/README.md
```diff
@@ -1,5 +1,12 @@
 # Cinema 2025 conda build recipe
 
+## Instructions for Cinema4D plugin packages
+Plugins can be placed into the Cinema4D's plugin directory. eg. `"$PREFIX/cinema4d/bin/plugins"` or `%PREFIX%\cinema4d\plugins`. This repository contains examples:
```
karthikbekalp: I think we should just point to the conda package recipe. This is because Cinema 4D plugins generally have their own set of requirements and some of them would not work out of the box if its just placed in the plugins folder. What do you think? (Not a blocker)

conda_recipes/cinema4d-2025/README.md
```diff
@@ -1,5 +1,12 @@
 # Cinema 2025 conda build recipe
 
+## Instructions for Cinema4D plugin packages
```
karthikbekalp: I wonder if we should place this section at the bottom similar to the AE instructions as these are optional steps and not required steps for the recipe? https://github.com/aws-deadline/deadline-cloud-samples/pull/131/files

karthikbekalp (approved):

lucaseck (approved):

PR #143, opened 2025-10-16
chore: Added a Blender plugin bundle sample.

erico-aws:

There is no example recipe that installs multiple Blender addons.

Created a example recipe that accepts an arbitrary number of Blender addons and installs all of them. Likewise, it removes them when the conda environment is deactivated. 

It follows the same pattern as the FLIP Fluids example.

Example for folks to see how to bring multiple plugins in a single conda package.

Used the submit package job script to submit and build conda package jobs in my own Deadline Cloud Farm. Verified that multiple plugins were installed.

I've added links to the Blender 4.5 sample readme to both this example and the FLIP Fluid sample.

---


conda_recipes/blender-plugin-bundle/recipe/build.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/sh
+set -xeuo pipefail
+
+INSTALL_DIR="$PREFIX/blender-plugins"
+mkdir -p $INSTALL_DIR
```
jusiskin: we should quote this to avoid bash accidentally treating this as two arguments:


```suggestion
mkdir -p "$INSTALL_DIR"
```

conda_recipes/blender-plugin-bundle/recipe/build.sh
```diff
@@ -0,0 +1,72 @@
...
+
+# All the addons are moved into the $INSTALL_DIR
+mv $SRC_DIR/plugins/* $INSTALL_DIR
```
jusiskin: same here:


```suggestion
mv "$SRC_DIR"/plugins/* "$INSTALL_DIR"
```

jusiskin (changes requested):
Looks good. Just two small shell quoting fixes

jusiskin (approved):

conda_recipes/blender-plugin-bundle/recipe/build.sh
```diff
@@ -0,0 +1,72 @@
...
+
+plugin_dir = r"$INSTALL_DIR"
+zip_files = glob.glob(os.path.join(plugin_dir, "*.zip"))
+
+print(f"Uninstalling {len(zip_files)} addons")
+for zip_file in zip_files:
+    # Extract addon name from zip file to remove it
+    with zipfile.ZipFile(zip_file, 'r') as z:
+        for name in z.namelist():
+            if name.endswith('__init__.py'):
```
lucaseck, line 49: Is there an easier way that on activate the list of addon names can be grabbed as they are enabled and then persist that somewhere in the env to use during deactivation instead of trying to find the names based on the directory structures etc?

lucaseck: Mainly was hoping that either the Blender API returned something pointing to the installed addon as the return value we could query. Other option I was thinking of was could you take a before and after list of all addons, then query the names out of that diff and persist them somewhere in the env.

erico-aws: It seems like, Blender names the plugins after the top level folder in the zip. I think the only other way of doing this would be to do the addon diff.

lucaseck (approved):

conda_recipes/blender-plugin-bundle/recipe/build.sh
lucaseck: Not going to block over it. If we find it's fragile can always change it.

PR #144, opened 2025-10-23
feat: Update conda package samples to use rattler S3 support

mwiebe:

The rattler library for working with conda packages and environments has merged and released https://github.com/conda/rattler/pull/1629 supporting the AWS SDK for default credentials when accessing S3. This means the sample conda package building can directly use rattler-build and rattler-index with S3 conda channels, no longer needing the mountpoint-s3 proxy.

This changes updates the conda package building scripts and samples to use that support.

* Updated and simplified the job bundle for building a conda recipe.
  * No longer use proxy channels based on mountpoint-s3 when using rattler. Now it directly passes the s3:// channel URLs.
  * Remove the separate channel indexing step, and instead call rattler-index after the package build. Since 0.26, rattler-index protects concurrent indexing operations.
* Remove the default of using the conda-build tool by making the buildTool field in deadline-cloud.yaml mandatory.
* Convert many sample recipes to use just rattler-build.
* Adjust the Python versions in the Python package sample recipes to be based on Python 3.13. The structure is flexible to build with more Python versions.
* Edit the README.md to Deprecate the conda-build support in the sample, for removal in a future update.
* Update the blender-flipfluids sample to not require nesting zip files. This needed a change in the package build script to keep the `url` field in `source` instead of converting it to `path`.

The job bundle for building packages is much simpler. Re-indexing a channel is a single command included in the package build instead of as a separate step. Windows packages can be built with rattler-build instead of conda-build.

Switched a number of conda-build sample recipes to rattler-build, and confirmed they build. Built a wide variety of packages and ran test jobs with them.

The README is updated to indicate that conda-build support is deprecated, and to emphasize rattler-build.

---


crowecawcaw (approved):
🙌

leongdl (approved):

PR #145, opened 2025-10-25
refactor: Add conda queue env samples with py-rattler

mwiebe:

The [rattler library](https://github.com/conda/rattler#readme) now supports the default AWS credentials configuration, so can be used to provide conda environments just like the samples we have using the conda command.

In the progress of testing, found some small tweaks necessary in other sample jobs.

Implement a conda queue environment that uses the [py-rattler library](https://conda.github.io/rattler/py-rattler/). A lot of it was straightforward, but the library doesn't use the default configuration that tools like `pixi` apply, so the code had to manually read `.condarc` configuration and enable S3 channel support.

* Tested with a variety of the job bundle samples, and updated them.
  Some updates, like adding 'pip' to CondaPackages, are for py-rattler,
  and others are to increase the clarity and reliability of the samples.
  * Change the control for CondaPackages parameters to HIDDEN. If
    there's a conda queue environment, it will show a GUI for this
    parameter, and if there isn't a conda queue environment, the
    parameter is unused so it's better to hide it.
  * Update the stage 4 job dev progression sample to include 'pip' in
    CondaPackages for compatibility with rattler env creation.
  * Update the cli_job sample so that it works when the input/output
    directory doesn't exist.
  * Update the afterfx_render_onetask sample to format the parameter
    grouping in the template, set the default package to 'aftereffects',
    set the default host requirements to specify windows, and add an
    output filename as the aerender command needs for '-output'. Switch
    to running a Python script that ensures the output directory exists.
  * Update copy_s3_prefix_to_job_attachments with host requirements for
    linux, and add 'pip' to CondaPackages.
  * Update CondaPackages for gsplat_pipeline to use 'colmap' instead of
    'colmap=*=gpu*'. Py-rattler gave a message like
    'exceptions.InvalidMatchSpecException: the build string '=gpu*' is
    not valid', and conda no longer needs this workaround.
  * Update keyshot_standalone to include keyshot in the default
    CondaPackages and to use a windows host requirement.

Another option for providing conda environments to Deadline Cloud jobs. Because the result often runs faster, we may want to make this the default approach after sufficient validation.

I performed some measurements of environment creation of this new queue environment compared to [conda_queue_env_console_equivalent.yaml](https://github.com/aws-deadline/deadline-cloud-samples/blob/mainline/queue_environments/conda_queue_env_console_equivalent.yaml). The job I used is [turntable_with_maya_arnold](https://github.com/aws-deadline/deadline-cloud-samples/tree/mainline/job_bundles/turntable_with_maya_arnold), and I fixed the test to use identical instance types.

When running the first time:
```
Conda:  1:28.22
Pyrattler:  55.70
```

When running again, using the package cache from the previous time:
```
Conda: 20.73
Pyrattler: 4.98
```

Configured the queue environment on a queue that has both Linux and Windows associated fleets. Ran a variety of jobs.

Added a description to the README.md for the queue environments.

---

### DIFF
--- a/queue_environments/README.md
+++ b/queue_environments/README.md
@@ -119,6 +119,19 @@ repeatedly re-download the same applications, but each session will have the ove
 packages into the virtual environment. Look at the sample with improved caching to reuse virtual
 environments across multiple jobs.
 
+### Conda queue environment using the py-rattler library
+
+The file [conda_queue_env_pyrattler.yaml](conda_queue_env_pyrattler.yaml) provides the same functionality as
+the above Conda queue environments, but uses the [py-rattler library](https://conda.github.io/rattler/py-rattler/).
+Rattler is a library that provides common functionality used within the conda ecosystem. It's written
+in Rust and tries to provide a clean API to its functionalities. The environments it creates are almost the same,
+but we found that py-rattler does not include 'pip' alongside 'python' by default, so if you need pip you must
+add it explicitly. It also raises an error for a subset of syntax that conda accepts, such as 'colmap=*=gpu*'.
+
+Testing has shown that this queue environment generally runs faster than the above when on the same instance types,
+but the error messages it produces when failing to solve for a virtual environment do not include as
+much detail to help diagnose what happened.
+
 ### Rez queue environment
 
 The file [rez_queue_env.yaml](rez_queue_env.yaml) provides the same functionality as
--- a/queue_environments/conda_queue_env_pyrattler.yaml
+++ b/queue_environments/conda_queue_env_pyrattler.yaml
@@ -1,0 +1,251 @@
+specificationVersion: "environment-2023-09"
+parameterDefinitions:
+- name: CondaPackages
+  type: STRING
+  description: >
+    This is a space-separated list of Conda package match specifications to
+    install for the job. E.g. "blender=4.5" for a job that renders frames in
+    Blender 4.5.
+  default: ""
+  userInterface:
+    control: LINE_EDIT
+    label: Conda Packages
+- name: CondaChannels
+  type: STRING
+  description: >
+    This is a space-separated list of Conda channels from which to install
+    packages. Deadline Cloud SMF packages are installed from the
+    "deadline-cloud" channel that is configured by Deadline Cloud.
+
+    Add "conda-forge" to get packages from the https://conda-forge.org/
+    community, and "defaults" to get packages from Anaconda Inc (make sure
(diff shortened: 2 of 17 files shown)


queue_environments/README.md
```diff
@@ -119,6 +119,17 @@ repeatedly re-download the same applications, but each session will have the ove
 packages into the virtual environment. Look at the sample with improved caching to reuse virtual
 environments across multiple jobs.
 
+### Conda queue environment using the py-rattler library
+
+The file [conda_queue_env_pyrattler.yaml](conda_queue_env_pyrattler.yaml) provides the same functionality as
```
baxeaz, line 124: This is great if this more or less is just a drop in performance improvement!  I guess there's no special licensing or security/configuration issues to worry about?  Licensing I see appears to be BSD-3 clause (And maybe just falls under Conda anyway?) and the configuration from what I can see briefly at https://conda.github.io/rattler/py-rattler/ largely aims to be a fairly simple from a configuration standpoint.

queue_environments/conda_queue_env_pyrattler.yaml
```diff
@@ -0,0 +1,237 @@
...
+            for channel in channels:
+                print(f"  - {channel}")
+            print(f"specs:")
+            for spec in specs:
+                print(f"  - {spec}")
+            print()
+
+            print("Solving the environment...")
+            # This list is copied from https://github.com/conda/rattler/blob/7d36fe334c14ddf6227dc0ec0982379d062c5915/crates/rattler-bin/src/commands/create.rs#L162-L172
+            middlewares = [
```
baxeaz, line 163: This is a bit odd that this is all exposed like this - You didn't come across any links describing why we'd need what more or less seems like a fair amount of boilerplate here?  We're adding S3 I see which I guess isn't default, and maybe the answer is "We didn't want to load a bunch of extra binaries that weren't going to be used" - I'd kind of expect like an Enum param or something still.

crowecawcaw (dismissed):

queue_environments/README.md
mwiebe: Yeah, that's right!

queue_environments/conda_queue_env_pyrattler.yaml
mwiebe: I couldn't find a way to do this, I think it would need changes to the pyrattler code.

baxeaz (approved):

crowecawcaw (approved):

PR #149, opened 2025-11-12
feat: Add upload_to_job_attachments utility script

shidarin:
Add utility_scripts directory and README
Update TLD README

Users would like some way of easily pre-populating large job attachment datasets into S3 from on-prem or workstations. While we had an S3 to S3 example, we lacked a standalone example.

Wrote a new python script which will help users upload various datasets to job attachments

New utility_scripts folder

Various AWS SAs have tested this script file over the last few weeks

Sidecar README

---

### DIFF
--- a/utility_scripts/upload_to_job_attachments/README.md
+++ b/utility_scripts/upload_to_job_attachments/README.md
@@ -1,0 +1,340 @@
+# AWS Deadline Cloud Job Attachments Uploader
+
+Upload files and directories from your local workstation to AWS Deadline Cloud job attachments storage.
+
+## Overview
+
+This script uploads files to the job attachments S3 bucket in content-addressable storage format, allowing subsequent jobs to use the data without re-uploading. It's useful for pre-populating job attachments with large datasets.
+
+## Features
+
+- Upload individual files or entire directories recursively
+- Multi-threaded uploads for better performance
+- Automatic deduplication (skips files already in S3)
+- Progress reporting and statistics
+- Retry logic with exponential backoff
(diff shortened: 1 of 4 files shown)


mwiebe (approved):
Looks good! I think we should integrate features of this into the deadline CLI command like the concurrency/bandwidth control, ability to use creds other than the queue IAM role for S3 access, and having a command to upload specified files and directories to job attachments as a pre-caching mechanism. We can approach that incrementally one at a time.

shidarin:
Make no mistake, a lot of the backend functionality for this is already in the CLI!

I chose to do a barebones implementation without the deadline library so it could serve as a sample for script and app usage, where `deadline-cloud`'s larger package and dependency list might be a factor.

utility_scripts/upload_to_job_attachments/README.md
```diff
@@ -0,0 +1,340 @@
+# AWS Deadline Cloud Job Attachments Uploader
```
leongdl, line 1: Thanks for creating this, the code looks correct, but maybe harder to maintain long term. 

In essence, this API does almost the same thing as this beta API:
https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/cli/_groups/manifest_group.py

https://github.com/aws-deadline/deadline-cloud/blob/mainline/src/deadline/client/cli/_groups/attachment_group.py

Can we finish / polish the Deadline CLIs ?

shidarin:
Updated to remove the manifest and tested, works just like before

PR #153, opened 2025-11-13
feat: add Houdini 21.0 Conda package sample recipe

lucaseck:
The only Houdini sample we had was for 20.5. 21.0 is out now and we should show how to build a Conda package for it.
Added a sample recipe, it was almst entirely just a copy/paste of the 20.5 recipe and changing version numbers/hashes. The one new addition is libxkbfile is now required to run `houdini` and `husk` directly it seems so that has been added to the dependencies being pulled in through dnf.

More samples for customers.

- Built the package in my own build farm following our public docs on how to set one up. Both of the `houdini -h` and `hython -h` tests passed. Did a manual render with Karma and a standalone husk submission which both succeeded.

- It has an updated README

---

### DIFF
--- a/conda_recipes/houdini-21.0/README.md
+++ b/conda_recipes/houdini-21.0/README.md
@@ -1,0 +1,166 @@
+# Houdini 21.0 Conda Recipe for AWS Deadline Cloud
+
+## Overview
+
+This directory contains a conda build recipe for Houdini 21.0.440, specifically configured for use with AWS Deadline Cloud. This package enables you to run Houdini rendering and processing jobs on Deadline Cloud service-managed fleets.
+
+## Package Information
+
+- **Application**: Houdini 21.0.440
+- **Supported Platforms**: linux-64
+- **Source**: SideFX Houdini downloads page
+- **License**: SideFXEULA
+- **Build Tool**: rattler-build
+
+## Prerequisites
+
+Before building this package, ensure you have:
+
+1. **AWS Deadline Cloud infrastructure** set up with:
+   - A farm configured for package building. See https://github.com/aws-deadline/deadline-cloud-samples/tree/mainline/cloudformation/farm_templates/starter_farm for instructions to create a Farm.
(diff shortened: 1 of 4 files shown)


conda_recipes/houdini-21.0/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Version Update Checklist
+
+To adapt this recipe for Houdini 20.5, 20.0 or 19.5:
+
+1. **Update Version Information**
+   ```yaml
+   # In recipe/recipe.yaml
+   context:
+     version: "20.0.xxx"  # or desired version
```
miabatta, line 113: Should we change this to 21.0.xxx for consistency?

lucaseck: This is the example of how to change to a previous version, that's why 20.0 is used.

miabatta: Gotcha, I misunderstood. Thanks!

miabatta (approved):

conda_recipes/houdini-21.0/deadline-cloud.yaml
```diff
@@ -0,0 +1,10 @@
+condaPlatforms:
+  - platform: linux-64
+    defaultSubmit: true
+    sourceArchiveFilename: houdini-21.0.440-linux_x86_64_gcc11.2.tar.gz
+    sourceDownloadInstructions: Download the installer archive from SideFX Houdini's downloads page.
```
erico-aws, line 5: Should we specify that you need to download the "old" installer and not the launcher download?

epmog (approved):

conda_recipes/houdini-21.0/deadline-cloud.yaml
lucaseck: I have a Redshift update next, so I'll add those updates to that. Good callout

PR #155, opened 2025-11-17
feat: add sample bundles for training a LoRA and generating images fr…

crowecawcaw:
We want to show how to train LoRAs on Deadline.

Add a pair of bundles - one for LoRA training and one for generating images from a LoRA. Also add instructions for using Pokemon as a dataset.

Cool demo!

I instructed Q to follow the instructions.

Yes
---


job_bundles/lora_image_diffusion/image_generation/template.yaml
```diff
@@ -0,0 +1,174 @@
...
+          --num-inference-steps {{Param.NumInferenceSteps}} \
+          --guidance-scale {{Param.GuidanceScale}} \
+          --seed {{Param.Seed}} \
+          --image-index {{Task.Param.ImageIndex}} \
+          --output-dir "{{Param.OutputDir}}" \
+          --lora-rank {{Param.LoRARank}}
+  hostRequirements:
+    attributes:
+    - name: attr.worker.os.family
+      anyOf: [linux]
```
mwiebe, line 168: Do these templates require a GPU? If so, you can add a min gpu amount of 1 like in https://github.com/aws-deadline/deadline-cloud-samples/blob/mainline/job_bundles/gsplat_pipeline/template.yaml#L249.

crowecawcaw: Not strictly necessary though it does run much faster with a GPU. I'll add a GPU requirement.

evanspearman-a (dismissed):

job_bundles/diffusion/image_generation/generate_image.py
```diff
@@ -0,0 +1,90 @@
...
+parser.add_argument("--seed", type=int, required=True)
+parser.add_argument("--image-index", type=int, required=True)
+parser.add_argument("--output-dir", required=True)
+parser.add_argument("--lora-rank", type=int, required=True)
+args = parser.parse_args()
+
+if not args.lora_path or not os.path.exists(args.lora_path):
+    raise ValueError(f"LoRA weights file is required but not found at: {args.lora_path}")
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
```
mwiebe: I've seen accelerator-independent code with pytorch before, and tracked some down in https://docs.pytorch.org/tutorials/beginner/basics/quickstart_tutorial.html#creating-models

```python
device = torch.accelerator.current_accelerator().type if torch.accelerator.is_available() else "cpu"
print(f"Using {device} device")
```

PR #166, opened 2025-12-19
feat: add sample for VPC resource endpoints

crowecawcaw:
Deadline Cloud supports connecting SMF to VPCs. The setup is a bit tricky.

Add a sample CloudFormation template and a test job to validate it as an SMF VPC starting point.

Helps people setting up up SMF with VPCs.

Ran through the instructions, submitted the job, verified the test file reads and writes correctly.

Yes
---

### DIFF
--- a/cloudformation/farm_templates/smf_vpc_fsx/deadline-cloud-smf-vpc-fsx-template.yaml
+++ b/cloudformation/farm_templates/smf_vpc_fsx/deadline-cloud-smf-vpc-fsx-template.yaml
@@ -1,0 +1,362 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: >
+  AWS Deadline Cloud service-managed fleet with FSx for OpenZFS storage accessed via VPC resource endpoint.
+  This template demonstrates how to connect Deadline Cloud workers to private VPC resources using VPC Lattice.
+
+  See the README.md included with the CloudFormation template to learn more.
+
+Parameters:
+  FarmName:
+    Type: String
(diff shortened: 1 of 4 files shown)


cloudformation/farm_templates/smf_vpc_fsx/deadline-cloud-smf-vpc-fsx-template.yaml
```diff
@@ -0,0 +1,362 @@
...
+        DeploymentType: SINGLE_AZ_1
+        ThroughputCapacity: 64
+        RootVolumeConfiguration:
+          NfsExports:
+            - ClientConfigurations:
+                - Clients: "*"
+                  Options:
+                    - rw
+                    - crossmnt
+                    - insecure
```
jaikish, line 170: minor: Shall we put a comment here why this(`insecure` option)  is required? Without this, NFS does not allow traffic from a set of ports which messes up the NATing layer.

https://security.stackexchange.com/questions/246527/what-is-insecure-about-the-insecure-option-of-nfs-exports

jaikish (dismissed):

cloudformation/farm_templates/smf_vpc_fsx/deadline-cloud-smf-vpc-fsx-template.yaml
crowecawcaw: Great call! Added a quick note about what it does.

jaikish (approved):

Cherie-Chen (approved):

PR #170, opened 2026-01-06
feat: Adds a worker configuration script to install fonts on Linux SMF

jblagden:

I needed to use fonts that aren't included on the service managed fleet image.

Download the font(s) from an S3 bucket and install them. This needs administrator powers, so I used a worker configuration script.

If you need a font on a Linux SMF you'll be able to use it. Assuming the font renders as expected, which isn't guaranteed.

Submitted a Nuke job that uses the font 'Arial' in and saw the text in the output image.

output

[NukeTextTest.nk.zip](https://github.com/user-attachments/files/24458674/NukeTextTest.nk.zip)

Yes, included a README.md.

---

### DIFF
--- a/host_configuration_scripts/linux_font_installation/README.md
+++ b/host_configuration_scripts/linux_font_installation/README.md
@@ -1,0 +1,58 @@
+# AWS Deadline Cloud Font Installation
+
+This script installs fonts from an S3 bucket on AWS Deadline Cloud service managed fleet instances, making them available to applications like Nuke.
+
+Be aware that fonts may render differently from operating system to operating system.
+
+## Setup
+
+### 1. Upload Fonts to S3
+Upload your font files to an S3 bucket:
+```bash
+aws s3 cp /path/to/your/fonts/ s3://your-bucket-name/Fonts/ --recursive
+```
+
+Using a separate folder for fonts is optional; beware this script will copy over everything in that folder. Making sure only fonts are getting moved will save time at instance startup.
+
+### 2. Configure the Script
+Edit `font_install_host_config.sh` and update these variables:
+```bash
+S3_FONTS_URI="s3://your-bucket-name/Fonts/"
+JOB_USER="job-user"  # This is job-user by default, update if needed
+```
+
+### 3. Add IAM Policy to Fleet
+Your Deadline Cloud fleet needs S3 read permissions. Add this policy to your fleet's IAM role:
+
+```json
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+                "s3:GetObject",
+                "s3:ListBucket"
+            ],
+            "Resource": [
+                "arn:aws:s3:::your-bucket-name",
+                "arn:aws:s3:::your-bucket-name/Fonts/*"
+            ]
+        }
+    ]
+}
+```
+
+Replace `your-bucket-name` with your actual S3 bucket name.
+
+## Usage
+
+Deploy the script through the AWS Deadline Cloud console:
+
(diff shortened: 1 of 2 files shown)


host_configuration_scripts/linux_font_installation/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+```json
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+                "s3:GetObject",
+                "s3:ListBucket"
```
crowecawcaw, line 35: Should this be ListObjects?

host_configuration_scripts/linux_font_installation/README.md
```diff
@@ -0,0 +1,58 @@
+# AWS Deadline Cloud Font Installation
+
+This script installs fonts from an S3 bucket on AWS Deadline Cloud service managed fleet instances, making them available to applications like Nuke.
+
+Be aware that fonts may render differently from operating system to operating system.
```
crowecawcaw, line 5: This is Linux only right? If so, let's note that.

jblagden: Yes, good catch!

host_configuration_scripts/linux_font_installation/README.md
jblagden: I don't follow - would that be a smaller scope of permissions?

host_configuration_scripts/linux_font_installation/font_install_host_config.sh
```diff
@@ -0,0 +1,45 @@
...
+echo "Installing fontconfig..."
+yum install -y fontconfig
+
+# Create fonts directory
+echo "Creating fonts directory..."
+mkdir -p "/home/$JOB_USER/.fonts"
+
+# Download fonts with error checking
+echo "Downloading fonts from S3..."
+if aws s3 cp "$S3_FONTS_URI" "/home/$JOB_USER/.fonts" --recursive; then
```
leongdl, line 22: Do we need to mention the Fleet role needs access to this bucket?

leongdl: Ah I see it in the README below, all good, but worth adding a line here in case users do not read the manual.

jblagden: Added that callout to the error handling so it shows up in worker logs if it fails.

leongdl (approved):

host_configuration_scripts/linux_font_installation/font_install_host_config.sh
```diff
@@ -0,0 +1,45 @@
...
+S3_FONTS_URI="s3://your-bucket-name/Fonts/"
+JOB_USER="job-user"
+
+echo "[$(date)] Starting font installation..."
+echo "S3 URI: $S3_FONTS_URI"
+echo "Job User: $JOB_USER"
+
+# Install fontconfig (required for font management)
+echo "Installing fontconfig..."
+yum install -y fontconfig
```
joel-wong-aws, line 14: Should we use `dnf` instead of `yum`? It's installed by default on Amazon Linux 2023 and is the standard package manager there with more features

host_configuration_scripts/linux_font_installation/README.md
crowecawcaw: Looking at the host config script, I don't see a use for ListBuckets but the recursive copy looks like it will need ListObjects. Does this IAM policy have a typo?

joel-wong-aws (approved):

PR #176, opened 2026-02-06
feat: add disconnect UBL queue environment sample

joel-wong-aws:

Users who want to use Bring Your Own License (BYOL) with custom license servers may want to ensure that deadline Cloud Usage Based License (UBL) is disconnected.

Added a new queue environment sample `disconnect_ubl_queue_env.yaml` that unsets all Deadline Cloud UBL-related environment variables.

The queue environment uses an embedded bash script that outputs `openjd_unset_env` commands for each variable. Updated the queue_environments README.md to document this new sample.

Users can now easily configure their queues to never fallback to Deadline Cloud UBL by adding this queue environment with priority 0 (to run before other queue environments).

- Added this queue environment to my queue and verified that all licensing environment variables were unset.

Yes, added documentation to `queue_environments/README.md` explaining:
- What the queue environment does
- When to use it (BYOL scenarios)
- How to configure it (priority 0)

---

### DIFF
--- a/queue_environments/README.md
+++ b/queue_environments/README.md
@@ -149,3 +149,17 @@ The core enhancement of this queue environment is to use named Conda environment
 jobs. The default environment name uses the hash of the Conda channels and packages, or you can explicitly
 set the name in the job. It also includes a parameter for how long to use an environment without running a package
 update, so that most of the time it will take seconds to activate an environment that's being reused.
+
+### Disconnect UBL queue environment
+
+The file [disconnect_ubl_queue_env.yaml](disconnect_ubl_queue_env.yaml) unsets Deadline Cloud Usage Based
+License (UBL) environment variables. Use this queue environment if you want to turn off all connections to
+Deadline Cloud UBL for your queue and force the use of a custom license server (see the
+[Bring Your Own License documentation](https://docs.aws.amazon.com/deadline-cloud/latest/developerguide/smf-byol.html)).
+
+This queue environment should be run before any other queue environments (for example, by setting the priority to 0)
+so that connections to your custom floating licenses (such as RLM) in other queue environments are not
+accidentally removed.
+
+Please note that this is a sample, additional UBL environment variables may be
+added in the future.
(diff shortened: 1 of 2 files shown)


queue_environments/disconnect_ubl_queue_env.yaml
```diff
@@ -0,0 +1,39 @@
...
+      data: |
+        vars=(
+          ADSKFLEX_LICENSE_FILE
+          LUXION_LICENSE_FILE
+          PIXAR_LICENSE_FILE
+          SESI_LMHOST
+          VRAY_AUTH_CLIENT_SETTINGS
+          foundry_LICENSE
+          g_licenseServerRLM
+          redshift_LICENSE
```
viknith, line 37: Nit: would you want to unset VRAY_AUTH_CLIENT_FILE_PATH and VRAY_AUTH_CLIENT_SETTINGS too?

queue_environments/README.md
```diff
@@ -149,3 +149,14 @@ The core enhancement of this queue environment is to use named Conda environment
 jobs. The default environment name uses the hash of the Conda channels and packages, or you can explicitly
 set the name in the job. It also includes a parameter for how long to use an environment without running a package
 update, so that most of the time it will take seconds to activate an environment that's being reused.
+
+### Disconnect UBL queue environment
+
+The file [disconnect_ubl_queue_env.yaml](disconnect_ubl_queue_env.yaml) unsets Deadline Cloud Usage Based
+License (UBL) environment variables. Use this queue environment if you want to turn off all connections to
```
viknith, line 156: We could add more env variables in the future, can we include a note saying that this sample may not always be the full list of env variables to unset and to reference some docs on the full list?

queue_environments/disconnect_ubl_queue_env.yaml
joel-wong-aws: `VRAY_AUTH_CLIENT_SETTINGS` is set to `/null` on both Windows and Linux by default on SMF (so it won't use UBL), and `VRAY_AUTH_CLIENT_SETTINGS` is already unset in this PR.

queue_environments/README.md
joel-wong-aws: Added!

viknith (approved):

mwiebe (approved):

PR #181, opened 2026-02-23
chore: Convert sample job templates to use task chunking for After Effects, Blender, Houdini and KeyShot

Cherie-Chen:
chore: Convert sample job templates to use task chunking for After Effects, Blender, Houdini and KeyShot

Now we support [Task Chunking](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md), we should update our sample job templates to use it.

Convert sample job templates to use task chunking. In this PR, it converts sample job templates forAfter Effects, Blender, Houdini and KeyShot and more to come. 

User will have sample job templates using task chunking for the DCC they are working with.

I submitted job using these job templates and here is the results:

* After Effects

Screenshot 2026-02-23 at 12 02 21 PM

* Blender

Screenshot 2026-02-23 at 12 05 36 PM

* Houdini

Screenshot 2026-02-23 at 12 04 54 PM

* KeyShot

Screenshot 2026-02-23 at 12 03 25 PM

Yes, this PR also included doc changes

---

### DIFF
--- a/job_bundles/afterfx_render_one_task/README.md
+++ b/job_bundles/afterfx_render_one_task/README.md
@@ -1,21 +1,49 @@
-# After Effects Render - one task
+# After Effects Render with Contiguous Chunks
 
-## Use case for this job
+This job bundle renders After Effects compositions using aerender with the [Task Chunking](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md) extension for efficient parallel rendering.
 
-This is an After Effects job bundle that allows the user
-to submit a job that uses aerender to render a frame range
-as a single task. This means the entire workload will render
-on one worker as one command.
+## Features
 
-It accepts the following job parameters that modify the render:
-* Project file
-* Comp name
-* Input directory
-* Output directory
-* Start frame
-* End frame
+- **Task Chunking**: Uses `CHUNK[INT]` with contiguous frame ranges for parallel rendering across multiple workers
+- **Adaptive Chunking**: Optional target runtime allows the scheduler to adjust chunk sizes dynamically
 
-This job bundle expects a user to specify an input directory that
-contains all the file references that are required to render.
-Generally, the project file should be within this directory to
-ensure relative paths are preserved.
+## Parameters
+
+| Parameter | Description | Default |
+|-----------|-------------|---------|
+| Project file | After Effects project file (.aep, .aepx) | - |
+| Comp name | Composition to render | - |
+| Input directory | Directory containing input files | - |
+| Output directory | Render output directory | - |
+| Frames | Frame range (e.g., `0-100`) | `0-50` |
+| Chunk Size | Number of frames per chunk | `10` |
+| Target Runtime | Target seconds per chunk (0 to disable) | `180` |
+
+## Task Chunking
+
+This template uses the `TASK_CHUNKING` extension:
+
+```yaml
+extensions:
+  - TASK_CHUNKING
+
+steps:
+- name: RenderComp
+  parameterSpace:
+    taskParameterDefinitions:
+    - name: Frame
+      type: CHUNK[INT]
+      range: "{{Param.Frames}}"
+      chunks:
+        defaultTaskCount: "{{Param.ChunkSize}}"
+        targetRuntimeSeconds: "{{Param.TargetRuntime}}"
+        rangeConstraint: CONTIGUOUS
+```
+
+Each chunk expands to a contiguous range like `"0-9"` or `"10-19"`, which maps directly to aerender's `-s` (start) and `-e` (end) arguments.
+
+Reference: [After Effects Automated Rendering](https://helpx.adobe.com/after-effects/using/automated-rendering-network-rendering.html)
+
+## Usage
+
+This job bundle expects a user to specify an input directory that contains all the file references required to render. Generally, the project file should be within this directory to ensure relative paths are preserved.
--- a/job_bundles/afterfx_render_one_task/README.md
+++ b/job_bundles/afterfx_render_one_task/README.md
@@ -1,11 +1,10 @@
 # After Effects Render with Contiguous Chunks
 
-This job bundle renders After Effects compositions using aerender with the [Task Chunking](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md) extension for efficient parallel rendering.
+This job bundle renders After Effects compositions using aerender with the [Task Chunking](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md) extension.
 
-## Features
+## Task Chunking
 
-- **Task Chunking**: Uses `CHUNK[INT]` with contiguous frame ranges for parallel rendering across multiple workers
-- **Adaptive Chunking**: Optional target runtime allows the scheduler to adjust chunk sizes dynamically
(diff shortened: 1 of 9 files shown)


job_bundles/afterfx_render_one_task/template.yaml
```diff
@@ -55,32 +57,39 @@ parameterDefinitions:
...
 # Frame range
-- name: StartFrame
+- name: Frames
+  type: STRING
+  userInterface:
+    control: LINE_EDIT
+    label: Frames
+    groupLabel: Frame range
+  default: "0-50"
+  description: The frame range to render (e.g., 0-100).
```
diane-hj: power nit. I think we could match the numbers between default and description? :) 0-50.

job_bundles/blender_render/README.md
```diff
@@ -0,0 +1,46 @@
+# Blender Render
+
+This job bundle renders Blender scenes using the [Task Chunking](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md) extension with non-contiguous frame support for flexible parallel rendering.
+
+## Features
+
+- **Non-Contiguous Chunking**: Uses `CHUNK[INT]` with `rangeConstraint: NONCONTIGUOUS` for arbitrary frame sets
+- **Adaptive Chunking**: Optional target runtime allows the scheduler to adjust chunk sizes dynamically
+- **Flexible Frame Ranges**: Supports ranges, steps, and individual frames (e.g., `1-10,15,20-100:2`)
```
diane-hj: "Flexible Frame Ranges" feels a bit redundant with "Non-Contiguous Chunking" — the ability to handle arbitrary frame sets is what non-contiguous chunking is. The other two bullets (Non-Contiguous Chunking and Adaptive Chunking) are self-explanatory, but "Flexible Frame Ranges" overlaps. What do you think about folding it into the Non-Contiguous Chunking bullet? e.g., "Non-Contiguous Chunking: Uses CHUNK[INT] with rangeConstraint: NONCONTIGUOUS for arbitrary frame sets including ranges, steps, and individual frames (e.g., 1-10,15,20-100:2)"

job_bundles/blender_render/template.yaml
```diff
@@ -1,105 +1,142 @@
...
+    label: Chunk Size
+  default: 5
+  minValue: 1
+  description: Number of frames per chunk.
+- name: TargetRuntime
+  type: INT
+  userInterface:
+    control: SPIN_BOX
+    label: Target Runtime (Seconds)
+  default: 600
```
diane-hj, line 40: Nit. The default was 180 in the above README.md :) Should we match?

job_bundles/afterfx_render_one_task/template.yaml
Cherie-Chen: Nice catch. I have updated it.

job_bundles/blender_render/README.md
Cherie-Chen: Good point. I have dropped this bullet point.

job_bundles/blender_render/template.yaml
Cherie-Chen: In test, I noticed Blender sometimes takes minutes to render a task, with 180s target-runtime, I ended up getting 1 task per chunk, hence I bumped it higher. I have updated the read me to match this

job_bundles/afterfx_render_one_task/README.md
```diff
@@ -1,21 +1,49 @@
...
-contains all the file references that are required to render.
-Generally, the project file should be within this directory to
-ensure relative paths are preserved.
+# After Effects Render with Contiguous Chunks
+
+This job bundle renders After Effects compositions using aerender with the [Task Chunking](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md) extension for efficient parallel rendering.
+
+## Features
+
+- **Task Chunking**: Uses `CHUNK[INT]` with contiguous frame ranges for parallel rendering across multiple workers
```
mwiebe: This needs a small adjustment. The reason for chunking is to reduce overhead by running chunks of short-running tasks together instead of scheduling each one individually. Chunking reduces parallelism. Particularly our new adaptive chunking mode is about automatically load balancing work more efficiently.

job_bundles/afterfx_render_one_task/template.yaml
```diff
@@ -1,8 +1,10 @@
 specificationVersion: jobtemplate-2023-09
-name: After Effects Render - one task
+extensions:
+  - TASK_CHUNKING
+name: After Effects Render with Contiguous Chunks
 description: >
-  A simple job bundle that allows a user to select
-  a project file and enter a comp to render with aerender.
+  A job bundle that renders an After Effects composition using aerender
+  with task chunking for efficient parallel rendering.
```
mwiebe: This description is better than above, because it says "efficient parallel rendering" instead of just "parallel rendering".

job_bundles/afterfx_render_one_task/README.md
```diff
@@ -1,21 +1,49 @@
...
-* Start frame
-* End frame
-
-This job bundle expects a user to specify an input directory that
-contains all the file references that are required to render.
-Generally, the project file should be within this directory to
-ensure relative paths are preserved.
+# After Effects Render with Contiguous Chunks
+
+This job bundle renders After Effects compositions using aerender with the [Task Chunking](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md) extension for efficient parallel rendering.
```
mwiebe: I think a nice benefit of using the task chunking feature is that rendering job templates can standardize on a single Frames job parameter that supports both ranges like 1-50 and pickup frames like 5,7-9,32 where you just want to re-render selected frames, and chunking applies well to both. With the StartFrame/EndFrame approach, you can't do the pickup frames in a simple way.

job_bundles/afterfx_render_one_task/README.md
```diff
@@ -1,21 +1,49 @@
...
-* Output directory
-* Start frame
-* End frame
-
-This job bundle expects a user to specify an input directory that
-contains all the file references that are required to render.
-Generally, the project file should be within this directory to
-ensure relative paths are preserved.
+# After Effects Render with Contiguous Chunks
+
```
eherozhao, line 2: should we keep both one task and chunking to help the reader know they all work?

Cherie-Chen: "One task" in the original title was misleading, in my opinion. The old sample had **1 step** rendering frames 0–50, which produced **51 tasks**. The new sample preserves exactly the same frame range (0–50) and the same step and task count — nothing about what the job template *does* has changed. The only addition is the `TASK_CHUNKING` extension element in the template, which enables the chunking capability.

The title change was meant to more clearly reflect what this job template sample actually demonstrates.

job_bundles/afterfx_render_one_task/README.md
Cherie-Chen: I have updated README in task_chunking sample folder to call out the benefits of task chunking and revised files in this PR.

job_bundles/afterfx_render_one_task/template.yaml
```diff
@@ -89,37 +98,48 @@ parameterDefinitions:
...
         data: |
           import os
-          import sys
           import subprocess
 
-          # Ensure the output directory exists
-          os.makedirs(sys.argv[1], exist_ok=True)
+          # Parse contiguous chunk range (e.g., "0-50" or "10-10")
+          chunk = "{{Task.Param.Frame}}"
+          start_frame, end_frame = chunk.split("-")
```
crowecawcaw, line 127: Does this work if the chunk isn't contiguous?

PR #183, opened 2026-03-02
chore: Convert sample job templates to use task chunking for Nuke, Maya, VRED and 3dsMax/Vray

diane-hj:
Convert sample job templates to use task chunking for Nuke, Maya, VRED and 3dsMax/Vray

Now we support [Task Chunking](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md), we should update our sample job templates to use it.

Convert sample job templates to use task chunking. In this PR, it converts sample job templates for Nuke, Maya, VRED, and 3dsMax/Vray

User will have sample job templates using task chunking for the DCC they are working with.

I submitted job using these job templates and here is the results:

- 3dsMax/Vray
Screenshot 2026-03-02 at 12 02 31 PM

- VRED
Screenshot 2026-03-02 at 12 03 24 PM

- Nuke 
Screenshot 2026-03-02 at 12 04 21 PM

- Maya
Screenshot 2026-03-02 at 12 06 19 PM

Yes, this PR also included doc changes

---

### DIFF
--- a/job_bundles/3dsmax_vray_denoiser/README.md
+++ b/job_bundles/3dsmax_vray_denoiser/README.md
@@ -1,14 +1,41 @@
 # 3ds Max V-Ray Denoiser Example
 
-This job bundle demonstrates rendering 3ds Max scenes with V-Ray, including automatic VRIMG to EXR conversion with denoising preservation and intelligent frame chunking.
+This job bundle demonstrates rendering 3ds Max scenes with V-Ray, including automatic VRIMG to EXR conversion with denoising preservation, using the Open Job Description task chunking extension to reduce scheduling overhead for multi-frame rendering.
+
+## Task Chunking
+
+This job bundle uses the [Task Chunking](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md) extension with `rangeConstraint: CONTIGUOUS` to reduce scheduling overhead by grouping frames together into chunks.
+
+```yaml
+extensions:
+  - TASK_CHUNKING
+
+steps:
+- name: Render EXRs
+  parameterSpace:
+    taskParameterDefinitions:
+    - name: Frame
+      type: CHUNK[INT]
+      range: "{{Param.Frames}}"
+      chunks:
+        defaultTaskCount: "{{Param.ChunkSize}}"
+        targetRuntimeSeconds: "{{Param.TargetRuntime}}"
+        rangeConstraint: CONTIGUOUS
+```
+
+Each chunk expands to a contiguous range like `"0-4"` or `"5-9"`, which maps directly to `3dsmaxcmd.exe -start:N -end:N`.
+
+## Job summary
+
+This job bundle renders 3ds Max scenes with V-Ray, including automatic VRIMG to EXR conversion with denoising preservation. Task chunking amortizes 3ds Max's startup and scene loading time by rendering multiple frames per invocation.
 
 ## Features
 
-- **Smart Frame Chunking**: Automatically handles both contiguous ranges (e.g., `1-100`) and non-contiguous ranges (e.g., `1-5,10,15-20`)
+- **Task Chunking**: Uses `CHUNK[INT]` with contiguous frame ranges to reduce scheduling overhead by rendering multiple frames per task
+- **Adaptive Chunking**: Optional target runtime allows the scheduler to adjust chunk sizes dynamically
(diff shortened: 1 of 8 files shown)


job_bundles/3dsmax_vray_denoiser/README.md
```diff
@@ -20,20 +47,14 @@ This job bundle demonstrates rendering 3ds Max scenes with V-Ray, including auto
 ## Parameters
 
 - **Scene File**: 3ds Max scene file (.max) to render
-- **Frames**: Frame range specification (supports both `1-100` and `1,5,10-20` formats)
+- **Frames**: Frame range to render (e.g., `0-10`)
```
Cherie-Chen: Customer can still submit noncontiguous frames like `1,5,10-20` for contiguous rendering. Scheduler will break it up into contiguous chunks. 

We should keep the old statement here

job_bundles/3dsmax_vray_denoiser/README.md
```diff
@@ -1,14 +1,41 @@
 # 3ds Max V-Ray Denoiser Example
 
-This job bundle demonstrates rendering 3ds Max scenes with V-Ray, including automatic VRIMG to EXR conversion with denoising preservation and intelligent frame chunking.
+This job bundle demonstrates rendering 3ds Max scenes with V-Ray, including automatic VRIMG to EXR conversion with denoising preservation, using the Open Job Description task chunking extension for efficient multi-frame rendering.
+
+## Task Chunking
+
+This job bundle uses the [Task Chunking](https://github.com/OpenJobDescription/openjd-specifications/blob/mainline/rfcs/0001-task-chunking.md) extension with `rangeConstraint: CONTIGUOUS` for efficient parallel rendering.
```
Cherie-Chen: > efficient parallel rendering

One feedback I got from my other chunking sample PR was TaskChunking does not make it more efficient in parallel rendering, we already do that. We should both remove this from our READMEs :)

Cherie-Chen: FYI. This is the feedback I received: 
> This needs a small adjustment. The reason for chunking is to reduce overhead by running chunks of short-running tasks together instead of scheduling each one individually. Chunking reduces parallelism. Particularly our new adaptive chunking mode is about automatically load balancing work more efficiently.

job_bundles/3dsmax_vray_denoiser/README.md
```diff
@@ -1,14 +1,41 @@
 # 3ds Max V-Ray Denoiser Example
 
-This job bundle demonstrates rendering 3ds Max scenes with V-Ray, including automatic VRIMG to EXR conversion with denoising preservation and intelligent frame chunking.
+This job bundle demonstrates rendering 3ds Max scenes with V-Ray, including automatic VRIMG to EXR conversion with denoising preservation, using the Open Job Description task chunking extension for efficient multi-frame rendering.
+
+## Task Chunking
```
leongdl, line 5: We don't support chunking from the submitter, so I will be more careful here.

Can we only change examples that have submitters that support chunking?

leongdl (changes requested):

job_bundles/3dsmax_vray_denoiser/README.md
diane-hj: Thanks for catching this. Updated it. 🙏

job_bundles/3dsmax_vray_denoiser/README.md
diane-hj: Thank you! I updated the README files to have something like, ` reduce scheduling overhead by grouping frames into chunks.`.

job_bundles/3dsmax_vray_denoiser/README.md
diane-hj: Thanks for the feedback. My understanding is that these examples are intended to submit via Deadline CLIs. https://github.com/aws-deadline/deadline-cloud-samples?tab=readme-ov-file#job-bundle-samples Since these sample bundles don't go through the DCC submitters, they shouldn't be affected by whether the submitter plugins support chunking yet. Let me know if you have any concerns though — happy to discuss further!

leongdl: I think it would be better if we logged a corresponding ticket in 3dsmax, upgrade the submitter, then release the sample. Otherwise it is hard for any user to know how to use chunking for DCCs that we do not have the UX for

PR #188, opened 2026-03-13
feat: Add frame support to Arnold standalone render and add new 2-step bundle that exports Maya scene files for standalone rendering with Arnold.

rickrams:
Update arnold_standalone_render:
- Add Frames parameter with per-frame task distribution
- Add #### frame number substitution in input paths
- Replace OutputFileName with OutputFilePrefix for frame-numbered output
- Remove extraneous Message parameter
- Remove test artifacts

Add maya_arnold_ass_export_render:
- Two-step pipeline: export .ass from Maya, then render with kick
- Export step runs arnoldExportAss once for all frames (efficient)
- Render step fans out per-frame kick tasks across workers
- Job attachments syncs exported .ass files between steps
- Camera parameter for selecting the render camera
- Frame step support (e.g. 1-100:2)

Tested with multiple Maya scene files

Yes

---


job_bundles/maya_arnold_ass_export_render/template.yaml
```diff
@@ -0,0 +1,194 @@
...
+            '{{Param.MayaSceneFile}}' \
+            '{{Param.OutputDir}}/ass/{{Param.OutputFilePrefix}}' \
+            '{{Param.Frames}}' \
+            '{{Param.Camera}}'
+
+    - name: ExportScript
+      type: TEXT
+      filename: export_ass.py
+      data: |
+        import sys, re
```
leongdl, line 113: nit, would be nice if larger scripts become standalone files and not embedded scripts

job_bundles/maya_arnold_ass_export_render/template.yaml
```diff
@@ -0,0 +1,194 @@
...
+  script:
+    actions:
+      onRun:
+        command: bash
+        args: ['{{Task.File.Run}}']
+    embeddedFiles:
+    - name: Run
+      type: TEXT
+      data: |
+        set -xeuo pipefail
```
leongdl, line 165: Same as above.

epmog (approved):

leongdl (approved):

PR #195, opened 2026-03-26
chore: Fix up Maya 2026 recipes for local rattler-build usage

mwiebe:

I tested the Maya 2026 sample recipe with local rattler-build, and found it needed tweaking to work.

* Fix up the recipe, also switched it to use env_vars.d similar to the recent Blender recipe updates
* Updated the recipe to match the downloaded filename
* Also add an option in turntable_with_maya_arnold to allow watermark rendering when no license is available.
* Removed the reference to older rattler-build bug.

The recipe works locally and on build farms.

Tested the build on an Amazon Linux 2023 host, confirmed a local turntable render using it.

Tested the build to a Deadline Cloud build farm, confirmed a farm turntable render using it.

N/A

---

### DIFF
--- a/conda_recipes/maya-2026/recipe/build.sh
+++ b/conda_recipes/maya-2026/recipe/build.sh
@@ -28,7 +28,7 @@ ln -r -s "$INSTALL_DIR/bin/maya$MAYA_VERSION" "$INSTALL_DIR/bin/maya"
 mkdir -p "$SRC_DIR/download"
 cd "$SRC_DIR/download"
 dnf download --resolve -y freetype alsa-lib fontconfig harfbuzz libbrotli graphite2 \
-    libxkbfile xcb-util-cursor xcb-util-wm xcb-util-keysyms libxkbcommon-x11 \
+    libxkbfile xcb-util-wm xcb-util-keysyms libxkbcommon-x11 \
     libva libvdpau pciutils-libs
 
 for RPM_FILE in *.rpm; do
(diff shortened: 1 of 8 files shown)


conda_recipes/maya-2026/recipe/build.sh
```diff
@@ -28,7 +28,7 @@ ln -r -s "$INSTALL_DIR/bin/maya$MAYA_VERSION" "$INSTALL_DIR/bin/maya"
 mkdir -p "$SRC_DIR/download"
 cd "$SRC_DIR/download"
 dnf download --resolve -y freetype alsa-lib fontconfig harfbuzz libbrotli graphite2 \
-    libxkbfile xcb-util-cursor xcb-util-wm xcb-util-keysyms libxkbcommon-x11 \
```
leongdl, line 31: Interesting, I remember this was added for maya 2026 and was a new addition. Any reason why we are removing it? Is it in the system environment already?

conda_recipes/maya-2026/recipe/recipe.yaml
```diff
@@ -11,7 +11,7 @@ package:
   version: ${{ version }}
 
 source:
-  url: file://archive_files/Autodesk_Maya_{{ major_version }}_Linux_64bit.tgz
+  path: ../../archive_files/Autodesk_Maya_${{ major_version }}_ML_Linux_64bit.tgz
```
leongdl, line 14: nit, why ../../ ? Wouldn't that put the file at conda_recipies?

leongdl: never mind, see it in the readme below

conda_recipes/maya-mtoa-2026/recipe/build.sh
```diff
@@ -25,6 +25,11 @@ rm -r $PREFIX/$MTOA_ROOT/docs
 # to never use LD_LIBRARY_PATH in Conda environments.
 patchelf --add-rpath '$ORIGIN/../../maya2026/lib' $PREFIX/$MTOA_ROOT/bin/libai_renderview.so
 
+# Add RPATH for libraries in $MTOA_ROOT/procedurals to $MAYA_LOCATION/plug-ins/xgen/lib for libAdskSeExpr.so
+for file in "$PREFIX/$MTOA_ROOT/procedurals"/*.so; do
```
leongdl, line 29: This is interesting, what test case exposed this? We should add this to our test suite.

mwiebe: There was a warning about it, for a test case we'd need to make a scene that uses the xgen plugin

conda_recipes/maya-2026/recipe/build.sh
mwiebe: The package didn't exist, and the tests worked without it.

jairaws (approved):

epmog (approved):

conda_recipes/maya-2026/recipe/build.sh
leongdl: OK, will follow up with the team tomorrow, just worried something can fail.

conda_recipes/maya-mtoa-2026/recipe/build.sh
leongdl: Thanks, will add a backlog for this, otherwise it will be come an escaped defect next round.

leongdl (approved):

PR #81, opened 2025-02-08
feat: Implement an example CloudFormation for a custom Deadline Cloud Balancer

Ahuge:
UPDATE: See https://aws.amazon.com/about-aws/whats-new/2026/04/deadline-cloud-job-scheduling/ for the feature directly in Deadline Cloud.

### What is this?
This is an example of implementing a Lambda function that acts as a custom Balancer by modifying the `maxCurrentWorkers on a job.

### What was the problem/requirement? (What/Why)
We had an internal customer that wanted to modify how the default scheduler assigned workers to tasks. They wanted it to work in a similar way to the Deadline 10 Balancer options. 
### What was the solution? (How)
Implement a scheduled Lambda function that runs every minute and calculates the weighting and enforces it by setting `maxCurrentWorkers`
### How was this change tested?
This has been running in the customer account for just over a week now. The client reports it works as expected. 
No extensive additional testing has been done.

### Was this change documented?
New README and template section

Lambda code was initially written in collaboration with Michael Yuan (thanks!)
---

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

PR #196, opened 2026-03-31
feat: Add scheduled standby worker count CloudFormation template

xuetinp:

Customers want to reduce costs by only maintaining standby workers during business hours. Standby workers provide a warm pool that eliminates job start latency, but they incur charges even when idle. There was no sample showing how to schedule standby worker count changes on a Deadline Cloud fleet based on a time-of-day schedule.

Added a new CloudFormation template (fleet_standby_scheduling) that uses EventBridge Scheduler and a Lambda function to update a fleet's standbyWorkerCount on a cron schedule. By default, it sets standby workers to 2 at 8:00 AM UTC Mon–Fri and resets to 0 at 5:00 PM UTC.

Key design decisions:

-  The template takes FarmId and FleetId as parameters rather than creating the fleet itself, so it works as an add-on for fleets created via the console, CLI, or CloudFormation.
-  The Lambda calls GetFleet first to read the current configuration, then merges in the standbyWorkerCount change via UpdateFleet. This avoids overwriting other fleet settings.
-  Supports both service-managed and customer-managed fleets.
-   Schedule timezone is configurable (defaults to UTC).

New sample template only — no changes to existing templates. Adds:

-   cloudformation/farm_templates/fleet_standby_scheduling/deadline-cloud-standby-scheduling-template.yaml
-   cloudformation/farm_templates/fleet_standby_scheduling/README.md
-   Updated cloudformation/README.md with a link to the new template

Deployed the template to a test AWS account with one-time cron schedules to verify end-to-end behavior:

- Created a test farm and service-managed fleet.
- Deployed the stack with start schedule at 21:41 UTC and end schedule at 21:49 UTC.
- Verified the start schedule fired and set standbyWorkerCount to 2:

Note: The Lambda runtime's bundled boto3 may not yet include the autoScalingConfiguration API model. A troubleshooting section in the README documents how to add a boto3 Lambda layer to resolve this.

Yes. A README is included with:

-  Overview explaining the use case and that it supports both SMF and CMF fleets
-  Prerequisites
-  Deployment instructions (console and CLI)
-  Customization examples (different schedules, timezones)
-  Cleanup instructions
-  Troubleshooting section for the boto3 Lambda layer workaround

### DIFF
--- a/cloudformation/farm_templates/fleet_standby_scheduling/README.md
+++ b/cloudformation/farm_templates/fleet_standby_scheduling/README.md
@@ -1,0 +1,87 @@
+# Scheduled Standby Workers for Deadline Cloud Fleets
+
+## Overview
+
+This CloudFormation template schedules [standby worker count](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/auto-scaling-configuration.html)
+changes on a Deadline Cloud fleet based on a time schedule. Standby workers are a warm pool of idle workers
+that can start processing jobs immediately without waiting for new instances to launch. By scheduling
+standby workers only during business hours, you reduce job start latency when your team is active
+while avoiding the cost of idle workers overnight and on weekends.
+
+For example, with the default settings this template sets the standby worker count to 2 at 8:00 AM UTC
+Monday through Friday, and resets it to 0 at 5:00 PM UTC.
+
+This template works with any existing Deadline Cloud fleet — whether it was created via the AWS console,
+CLI, or another CloudFormation template. It supports both
+[service-managed fleets](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/smf-manage.html) and
+[customer-managed fleets](https://docs.aws.amazon.com/deadline-cloud/latest/userguide/manage-cmf.html).
+Use this sample as a starting point and modify it to fit your own scheduling and cost optimization needs.
+
+## Resources Created
+
+1. **Lambda Function** — Updates the fleet's standby worker count by calling the Deadline Cloud
+   `UpdateFleet` API. It reads the fleet's current configuration first to avoid overwriting other settings.
+2. **IAM Roles** — A Lambda execution role with `deadline:GetFleet` and `deadline:UpdateFleet` permissions
+   scoped to the specified fleet, and a scheduler role to invoke the Lambda.
+3. **EventBridge Schedules** — Two schedules that trigger the Lambda at the start and end of business hours.
+
+## Prerequisites
+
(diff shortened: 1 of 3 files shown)


npmacl (dismissed):

cloudformation/farm_templates/fleet_standby_scheduling/deadline-cloud-standby-scheduling-template.yaml
```diff
@@ -0,0 +1,200 @@
...
+  OffHoursStandbyWorkerCount:
+    Type: Number
+    Description: Standby worker count to set outside business hours.
+    Default: 0
+    MinValue: 0
+  BusinessHoursStartCron:
+    Type: String
+    Description: >
+      Cron expression for when business hours start (UTC).
+      Default is 8:00 AM UTC, Monday through Friday.
```
mwiebe: From this description I thought it was hardcoded to UTC, because the cron expression doesn't have a timezone in it. Then I saw there is a separate parameter for it. Maybe there's a way to make it more obvious?

"Default hours are 8:00 AM to 5:00 PM, and the default time zone is UTC."

cloudformation/farm_templates/fleet_standby_scheduling/README.md
```diff
@@ -0,0 +1,118 @@
...
+aws cloudformation create-stack \
+  --stack-name StandbyScheduling \
+  --template-body file://deadline-cloud-standby-scheduling-template.yaml \
+  --capabilities CAPABILITY_IAM \
+  --parameters \
+    ParameterKey=FarmId,ParameterValue=farm-0123456789abcdef0123456789abcdef \
+    ParameterKey=FleetId,ParameterValue=fleet-0123456789abcdef0123456789abcdef
+```
+
+### Add a boto3 Lambda layer
```
mwiebe: We came up with a way to avoid having to do this kind of thing in https://github.com/aws-deadline/deadline-cloud-samples/pull/81/changes#diff-4811d7fddad10b3bf73f9e4c7d3a7c328ec554e394901ec0ffda3869eb0648d9R40, what do you think of that approach?

xuetinp: Good suggestion! Will fix

cloudformation/farm_templates/fleet_standby_scheduling/deadline-cloud-standby-scheduling-template.yaml
xuetinp: Will revise the description!

mwiebe (approved):

npmacl (approved):

PR #204, opened 2026-04-20
feat: Add FFmpeg movie from job output sample job bundle

rickrams:
Add a new job bundle that downloads the output of a completed render job in the same queue and encodes the image sequence into an MP4 video using FFmpeg. Configurable parameters include frame rate, encoding preset, quality (CRF), pixel format, and resolution.

The job uses the deadline.job_attachments Python API to download outputs and FFmpeg from conda-forge for video encoding. A pre-submission hook injects the queue's S3 bucket settings so no additional IAM permissions are needed on the queue role. Image format is auto-detected from the downloaded files.

Given it uses hooks that is merged but not release will leave this in draft for now.

Wanted to be able to turn frames created in one jobo on the farm into a movie without having to download and re-upload the frames.

Submitted using a variety of test jobs with different frame formats and extensions

Sample has a readme.

---


job_bundles/ffmpeg_movie_from_job_output/hooks.yaml
```diff
@@ -0,0 +1,5 @@
+version: "1.0"
```
leongdl, line 1: Do we mean to check this in?

rickrams: yes! based on feedback on the hooks PR hooks files have a version - https://github.com/aws-deadline/deadline-cloud/blob/mainline/docs/submission-hooks.md

job_bundles/ffmpeg_movie_from_job_output/template.yaml
```diff
@@ -0,0 +1,282 @@
...
+            downloader.set_root_path(root, download_dir)
+        stats = downloader.download_job_output(
+            file_conflict_resolution=FileConflictResolution.OVERWRITE,
+        )
+        print(f"Downloaded {stats.processed_files} files ({stats.processed_bytes} bytes)")
+    - name: Run
+      type: TEXT
+      runnable: true
+      data: |
+        #!/bin/env bash
```
leongdl: Can we use only 1 programming language? EG all in python? A mixed shell + Python template is a bit odd.

And can we move the inline code to a file and we run the file? (Readability question)

leon-li-inspire (approved):

crowecawcaw (approved):

PR #214, opened 2026-05-08
feat: Add a Windows SSM to SMF worker example

leongdl:
Fixes: *<insert link to GitHub issue here>*

Deadline Cloud already ships a Linux sample (`job_bundles/ssh_to_smf`) that registers a
running worker as an SSM hybrid managed node so an operator can `aws ssm start-session`
into the host for the duration of a job. Customers debugging Windows render farms have
the same need — RDP / PowerShell / SSH into the live worker to inspect the job-user
environment, reproduce a failure, or tail logs — but there was no equivalent sample for
Windows, and the Linux bundle can't be reused because:

Added `job_bundles/ssh_to_smf_windows/` — a Windows sibling of `ssh_to_smf`. It consists
of:

- `setup/host_config.ps1` — Deadline Cloud fleet host-configuration script (runs as
  Administrator on every fresh worker). It provisions an `RDP` local admin user,
  enables RDP + the firewall rule, disables the UAC consent prompt, adds `job-user` to
  `Administrators`, and registers a SYSTEM-elevated scheduled task
  (`DeadlineSsmElevated`) whose DACL grants `job-user` `GenericRead + GenericExecute`.
  The task is the mechanism that lets the job trigger the `amazon-ssm-agent.exe
  -register` codepath without hitting UAC token filtering.
- `job/template.yaml` — the job template. It checks outbound reachability to the SSM
  endpoints, asks the elevated task to stop the agent, run `-register -clear`, run
  `-register -code ... -id ... -region ...` with a hybrid activation, reinstall the
  `AmazonSSMAgent` Windows service (the `-register` codepath removes it on this AMI
  flavor), and start the service. It extracts the `mi-*` managed node ID from the
  captured elevated-task output, prints it to the job log, and on exit deregisters the
  node and cleans up. `hostRequirements` pins execution to Windows workers.
- `submit.sh` / `submit.ps1` — bash and PowerShell submitters. Both call
  `ssm create-activation` with `--registration-limit 10` (Deadline retries failed jobs
  up to 5 times, each consuming one slot), then `deadline bundle submit` with the
  activation code/id passed in as job parameters.
- `README.md` — one-time account setup (SSM service role, advanced-instances tier,
  network), how to deploy the host config to a fleet via the console or
  `aws deadline update-fleet`, how to submit, how to connect (interactive PowerShell,
  RDP port-forward, SSH-over-SSM), how to log in as `job-user` after RDPing in, and a
  troubleshooting table covering the failure modes I hit while developing the bundle.

- New opt-in sample only. No existing bundles, host configs, or templates are modified.
- Users who deploy `setup/host_config.ps1` to a fleet are intentionally making workers
  in that fleet RDP-accessible to the `RDP` local admin account, which is a security
  trade-off. The README flags this explicitly and recommends shutting the fleet down
  when debugging is complete.

- Deployed `setup/host_config.ps1` to a service-managed Windows fleet in us-west-2 via
  `aws deadline update-fleet --host-configuration file://...`, cycled the fleet to 0
  and back to 1 worker, and confirmed from the `/aws/deadline/{farm}/{fleet}` worker
  log stream that the script completed with exit code 0, the `RDP` user was created,
  the `DeadlineSsmElevated` task was registered with the expected DACL, and
  `job-user` was in the Administrators group.
- Submitted the bundle with `./submit.sh farm-... queue-...`, watched the job log for
  `SSM Managed Node ID: mi-...`, and verified:
  - `aws ssm describe-instance-information` reported the mi-* node as `Online` within
    ~30 seconds.
  - `aws ssm start-session --target mi-...` gave an interactive PowerShell prompt.
  - `aws ssm start-session ... --document-name AWS-StartPortForwardingSession` plus
    `mstsc /v:localhost:13389` let me RDP in as the `RDP` admin.
  - After running `Set-LocalUser -Name job-user -Password (...)` in the RDP session,
    reconnecting as `job-user` worked and dropped me into the job-user profile.
  - On job exit, the node was deregistered (`aws ssm describe-instance-information`
    no longer returned it).
- Re-ran the host config against the same worker to confirm idempotency — the RDP
  password rotated, the scheduled task was left alone, and the script still exited 0.
- Also tested the PowerShell submitter (`submit.ps1`) from a Windows workstation to
  confirm parity with `submit.sh`.

Yes. The bundle ships with a full `README.md` covering:

- How the flow works end-to-end (activation → elevated task → mi-* node → session).
- Why the SYSTEM-elevated scheduled task is required (UAC token filtering).
- One-time account setup: SSM service role, advanced-instances tier, network egress.
- Two equivalent ways to deploy the host config (console vs.
  `aws deadline update-fleet`), including the fleet-cycle dance to pick up a new
  host config.
- Usage for both `submit.sh` and `submit.ps1`, including custom session duration,
  IAM role, and region.
- Three connection modes: interactive PowerShell, RDP via port-forward, and
  SSH-over-SSM.
- A subsection on how to log in as `job-user` via RDP (set a password with
  `Set-LocalUser` from the admin session first).
- A security note recommending shutting the fleet down after debugging so an
  RDP-capable worker is not left running.
- A troubleshooting table keyed on the actual error strings you see in the job log
  or host-config log, mapped to root causes and fixes.
- A cross-link to the Linux sibling (`ssh_to_smf`) and to the design doc under
  `.kiro/specs/ssh-to-smf-windows/design.md`.

---


job_bundles/ssh_to_smf_windows/submit.sh
```diff
@@ -0,0 +1,106 @@
...
+# Write a temp script to parse the activation JSON (avoid inline python3 -c)
+cat > /tmp/parse_activation.py << 'PYEOF'
+import json, sys
+data = json.load(sys.stdin)
+print(data["ActivationCode"])
+print(data["ActivationId"])
+PYEOF
+
+ACTIVATION_CODE=$(echo "$ACTIVATION" | python3 /tmp/parse_activation.py | sed -n '1p')
+ACTIVATION_ID=$(echo "$ACTIVATION" | python3 /tmp/parse_activation.py | sed -n '2p')
```
leon-li-inspire: Maybe we don't need to parse `parse_activation.py` twice  ? 

```
 read -r ACTIVATION_CODE ACTIVATION_ID < <(echo "$ACTIVATION" | python3 -c "
  import json, sys
  d = json.load(sys.stdin)
  print(d['ActivationCode'], d['ActivationId'])
  ")
```

job_bundles/ssh_to_smf_windows/job/template.yaml
```diff
@@ -0,0 +1,359 @@
...
+            "ssmmessages.$region.amazonaws.com",
+            "ec2messages.$region.amazonaws.com"
+        )
+        foreach ($endpoint in $endpoints) {
+            try {
+                $r = Test-NetConnection -ComputerName $endpoint -Port 443 -WarningAction SilentlyContinue
+                if ($r.TcpTestSucceeded) {
+                    Write-Host "  OK   ${endpoint}:443"
+                } else {
+                    Write-Host "  FAIL ${endpoint}:443 (would block hybrid registration)"
```
leon-li-inspire, line 158: Do we need to fail fast if the connection check failed?

job_bundles/ssh_to_smf_windows/job/template.yaml
```diff
@@ -0,0 +1,359 @@
...
+        # ------------------------------------------------------------------
+        # Cleanup / Deregister (best-effort; do not fail the job).
+        # ------------------------------------------------------------------
+        Write-Host ''
+        Write-Host 'Session ended. Deregistering managed node...'
+        $cleanupScript = Join-Path $elevatedDir 'cleanup.ps1'
+        @(
+            "`$ErrorActionPreference = 'Continue'",
+            "& '$agentExe' -register -clear 2>&1",
+            "Stop-Service -Name AmazonSSMAgent -Force -ErrorAction SilentlyContinue",
```
leon-li-inspire, line 356: Do we also need to remove the hybrid node from the SSM inventory after stopping service ?

job_bundles/ssh_to_smf_windows/job/template.yaml
leongdl: Sure, can change.

job_bundles/ssh_to_smf_windows/job/template.yaml
leongdl: Let me try, before I could not and then I see a bunch of `mi` instances in SSM. Have to check.

job_bundles/ssh_to_smf_windows/submit.sh
leongdl: Sure, easy enough.

job_bundles/ssh_to_smf_windows/job/template.yaml
leongdl: Done - updated.

job_bundles/ssh_to_smf_windows/job/template.yaml
leongdl: No we cannot - because the queue role here will not have SSM Activation / Deactivation credentials in the IAM role. We don't want jobs spawning managed nodes.

leon-li-inspire (approved):

job_bundles/ssh_to_smf_windows/submit.ps1
```diff
@@ -0,0 +1,109 @@
...
+Write-Host "Region:   $Region"
+Write-Host "Farm:     $FarmId"
+Write-Host "Queue:    $QueueId"
+Write-Host ''
+
+# --- Create SSM hybrid activation ---
+Write-Host 'Creating SSM hybrid activation...'
+$activationJson = aws ssm create-activation `
+    --iam-role $IamRole `
+    --registration-limit 1 `
```
seant-aws, line 68: i noticed in the read.me it says reg limit is 10  ?

seant-aws (approved):

job_bundles/ssh_to_smf_windows/submit.ps1
leongdl:
```
> **Registration limit:** `submit.sh` and `submit.ps1` both create an activation with `--registration-limit 10`. Deadline retries a failed job up to 5 times by default, and each retry consumes one registration count. Dropping this back to 1 is fine in production once the job is stable.
```

Yeah, its just a toggle parameter to allow jobs to re-run. The defaults are correct.

PR #216, opened 2026-05-08
feat: instructions on 3ds max installer zip creation

yuanmich2:
Added instructions on creating a .zip out of the 3ds Max installer files for the 3ds Max host configuration scripts and updated the host configuration scripts to have comments pointing to the documentation. I also ensured the file paths would actually match if you followed the new instructions.

The existing instructions were not very clear about how to create the installer .zip archive or why their format differed from what you download directly from the Autodesk website. Also, the host configuration scripts used variables like $3DS_MAX_FOLDER_NAME whose value had to be changed to make the script work, but whose correct value was not obvious and often required guessing and checking. The new instructions eliminate that.

Added additional instructions and screenshots to the main 3ds max host config readme. I then added references to those instructions throughout the other host config readmes as well as in comments in the host configuration scripts themselves.

I also reduced the total number of environment variables that needed to be replaced to make the scripts work and simplified some extraneous steps. For example, instead of downloading plugins to a temp folder, then making the plugins directory and copying to the plugins directory, I changed the scripts to create the plugins directory and download to there straight from the S3 bucket. There were a number of other small fixes like eliminating the hardcoded "resources" folder that needed to exist in the S3 bucket and more robust handling of files with different paths and filenames.

There clearer instructions and the scripts are more robust.

I tested locally on the EC2 instance where I was taking the screenshots. I have only tested against 3ds Max 2026 and the matching V-Ray version. I did not test my changes against all the plugins in the examples nor the other versions of 3ds Max and V-Ray.

The majority of the changes were edits the the README files in the 3ds max host configurations directory. I also provided lots of new comments in the host configuration scripts.

---


leongdl (approved):

godobyte (approved):

seant-aws (approved):

ZainAallii:
looking at it for future versions.

ZainAallii:
@godobyte @leongdl mentioning you guys. 

Currently our scripts do this:

```powershell
$BUCKET_NAME="your-bucket-name"
$3DS_MAX_ZIP_FILE="3ds-max-2025.zip"

# then later...
aws s3 cp "s3://$BUCKET_NAME/$3DS_MAX_ZIP_FILE" C:\3dsmax_setup\
```

PR #216 changes it to:

```powershell
$3DS_MAX_INSTALLER_ZIP_S3_URI="s3://your-bucket-name/path/to/3ds-max-2025.zip"

# then later...
aws s3 cp "$3DS_MAX_INSTALLER_ZIP_S3_URI" C:\3dsmax_setup\
```

Each installer gets its own full S3 URI variable instead of combining a shared bucket name with a filename. 
I am looking for your feedback: Do we want to change it for every add-on installed along with 3dsmax or just the the Max's installer?

yuanmich2:
I made that change to reduce the room for error. It wasn't clear that the files had to be in the root of the S3 bucket and the script would fail if you tried to put them in a folder/prefix. But with the new version, any path should work.

The only risk I see is if people try to construct the S3 URI themselves, then they could make more mistakes. But I'm hoping people just use the "Copy S3 URI" button that's built into the S3 interface to get the correct path every time.

PR #221, opened 2026-05-14
fix: Add memory host requirement to prevent OOM failures

anika-suman-amazon:
The FLUX.2 Klein 4B model requires ~8GB RAM to load transformer weights, causing OOM on instances with 32 GiB when combined with OS/CUDA overhead. Added amount.worker.memory min 65536 MiB to both job templates and updated README fleet requirements.

Jobs get scheduled on instances with 32 GiB RAM (e.g., g5.2xlarge), which isn't enough to load the 7.7 GB transformer weights plus PyTorch/CUDA overhead. This causes OOM failures during training.

Added amount.worker.memory: min 65536 (64 GiB) to hostRequirements in both job templates. Updated README fleet requirements to document this.

Jobs will only schedule on instances with 64 GiB+ RAM, preventing OOM failures.

- g5.2xlarge (32 GiB): OOM at model load and during training
- g5.4xlarge (64 GiB): Training and image generation both completed successfully

Yes--updated README fleet requirements for both job bundles.

---


job_bundles/flux2_klein_lora/README.md
```diff
@@ -25,6 +25,7 @@ Train custom LoRA adapters for FLUX.2 Klein models using your own image datasets
 
 **Fleet requirements:**
 - GPU with 13GB+ VRAM
+- 64 GiB+ system memory (the 4B model requires ~8GB RAM to load before GPU transfer)
```
Cherie-Chen: NIT: Consider removing "(the 4B model requires ~8GB RAM to load before GPU transfer)." As written, it raises more questions than it answers — if the model only needs ~8GB RAM to load, why is the requirement 64 GiB+? Without explaining what accounts for the remaining ~56GB (e.g., training overhead, batch processing, intermediate tensors), this detail may confuse readers rather than help them. A simpler line like - 64 GiB+ system memory keeps the README concise and avoids introducing unexplained context.

Cherie-Chen (approved):

job_bundles/flux2_klein_lora/README.md
anika-suman-amazon: Addressed!

leon-li-inspire (approved):

PR #239, opened 2026-06-12
docs: Shrink comments to fit 15000 char host config limit

viknith:

The After Effects persistent volume host config script exceeded the 15,000 character limit for host configuration scripts.

Trimmed verbose doc comments and decorative separator lines while preserving all runtime logging and functional code. Reduced from 15,505 to 14,700 characters.

No functional change. Only code comments were shortened.

- Verified character count is under 15,000 (wc -c reports 14,700).
- No logic was modified; script behavior is identical.

N/A — comments were shortened but the script's inline documentation remains intact.

---


crowecawcaw:
The original PR said the host config was tested. Was it not? Was this new host config tested?

viknith:
Looked back at the original script that was put up, and it was tested, but the comments that were added made it go over the limit, so I just stripped it out

viknith:
Doing another render test just in case

viknith:
Renders are passing, persistent volumes feature working as intended

crowecawcaw (approved):

MarinMaksutaj (approved):

andychoquette (approved):

PR #241, opened 2026-06-16
feat: Add cuda package and CondaChannels parameter to gsplat pipeline

anika-suman-amazon:
Add cuda to the default conda packages and introduce a new CondaChannels parameter so the queue environment can be configured with the required channels (deadline-cloud, conda-forge) from the job submission UI.

The gsplat pipeline job template did not include `cuda` in its default conda packages, users had to manually ensure their queue environment was set up with the correct channels (deadline-cloud, conda-forge) for package resolution.

- Added `cuda` to the default `CondaPackages` parameter value.
- Added a new `CondaChannels` parameter (STRING type, LINE_EDIT control) under the "Software Environment" group, with a default of `"deadline-cloud conda-forge"`. This follows the same pattern used in other job bundles (e.g., `monte_carlo_simulation`).

Users submitting the gsplat pipeline job will now have `cuda` included by default and can configure the conda channels directly from the submission UI without needing to manually edit their queue environment.

- Confirmed the new `CondaChannels` parameter appears in the correct UI group alongside `CondaPackages`.

---


Cherie-Chen (approved):

gianelli-lagos-aws (approved):
😃

PR #244, opened 2026-06-22
fix: ayon-launcher build script + cross-shell tests, document build-host requirement

leon-li-inspire:
## What

Two changes to the `ayon-launcher` recipe, plus a docs note:

1. Move the packaging logic out of the inline `build.script` lists into separate `build.sh` (linux) / `build_win.sh` (win), invoked with `bash`.
2. Replace the `tests` `script` blocks (`test -x/-f/-d`) with `package_contents` file globs.
3. Document in the recipe README and a build comment that each conda platform must be built on a matching-OS host/fleet.

## Why

A user reported the recipe **builds the linux target fine from Linux but fails when building from Windows**, erroring on "bad syntax" at the top of the unix build block.

I verified **all four host × target combinations** on EC2 (Linux + Windows workers, rattler-build 0.67):

| Build target | Build host | Before this PR | After this PR |
|---|---|---|---|
| linux-64 | Linux  | ✅ pass | ✅ pass |
| win-64   | Windows | ❌ build failed (inline bash under cmd.exe) | ✅ build passes; test now passes |
| linux-64 | Windows | ❌ fails | ❌ still unsupported (documented) |
| win-64   | Linux   | ❌ fails | ❌ still unsupported (documented) |

### Root cause

rattler-build runs the build script through the **build host's native shell** (`bash` on Linux, `cmd.exe` on Windows), while the `if: unix` / `if: win` selectors choose the branch by the **target** platform. These are independent axes:

- The original inline bash (`mkdir -p`, `cp -r`, heredocs) can't be parsed by `cmd.exe`, so building win-64 on a Windows worker failed. Moving the logic into `build.sh` / `build_win.sh` and invoking `bash` fixes that — **verified**: win-64 builds on Windows, linux-64 on Linux.
- **Cross-OS building cannot be fixed in-recipe**: on a Windows host `cmd.exe` won't expand `$RECIPE_DIR`; on a Linux host `bash` won't expand `%RECIPE_DIR%`. Each invocation fails with exit 127. This matches how `submit-package-job` already works — it runs rattler-build on a fleet worker whose OS matches the target platform. Documented in the README and the build comment rather than papered over.

### Test fix (found during verification)

The win-64 package **test** failed even on a Windows host: `test -x/-f/-d` is a posix builtin that `cmd.exe` can't run (exit 9009). Replaced with `package_contents` globs, which rattler-build evaluates directly against the built package **without launching a shell**, so they pass on both platforms.

## Testing

All combinations above were run on real EC2 instances via SSM. Recipe validated as well-formed YAML; `build.sh` / `build_win.sh` pass `bash -n`.

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/conda_recipes/ayon-launcher/README.md
+++ b/conda_recipes/ayon-launcher/README.md
@@ -26,6 +26,23 @@ The AYON Launcher conda package provides the **runtime environment**. The studio
 pixi global install rattler-build
 ```
 
+> **Build each platform on a matching-OS host.** The build scripts are bash
+> (`build.sh` / `build_win.sh`), and rattler-build runs the build script through the
+> **build host's native shell** — bash on Linux, `cmd.exe` on Windows — while the
+> `if: unix` / `if: win` selectors choose the branch by the **target** platform. As a
+> result:
+> - Build **linux-64 on a Linux host** (bash runs `build.sh` directly).
+> - Build **win-64 on a Windows host** (`cmd.exe` expands `%RECIPE_DIR%` and launches
+>   bash to run `build_win.sh`; bash ships with the Deadline Cloud sample Windows
+>   workers and Git for Windows).
+>
+> Cross-OS builds are **not supported** — e.g. building linux-64 from Windows fails
+> because `cmd.exe` does not expand `$RECIPE_DIR`, and building win-64 from Linux fails
+> because bash does not expand `%RECIPE_DIR%`. When submitting with
+> `submit-package-job`, the job runs rattler-build on a fleet worker, so ensure the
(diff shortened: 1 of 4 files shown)


crowecawcaw (dismissed):

leon-li-inspire:
Closing this PR. Verified on EC2 that native cross-OS rattler-build (e.g. building linux-64 on a Windows host) is a limitation of the shared recipe pattern used across all samples in this repo, not specific to ayon-launcher — the supported path is to build each conda platform on a matching-OS fleet via `submit-package-job`. Closing rather than merging while we reconsider the right approach.

PR #254, opened 2026-07-14
feat: Consolidate AE + Red Giant host config scripts and make it more robust

### LINKED ISSUE #243: Bug: After Effects Red Giant host config contains throw statements that obfuscates errors
### Describe the bug

If you write a throw statement into a host config script like this:

Image

It'll print out a generic "Powershell exited with 1" message. This makes it very hard to tell what actually broke.

### Expected Behaviour

We should let scripts fail normally or print better error messages.

### Current Behaviour

The AE+Red Giant host config script uses throw statements which will make it harder to debug than if you just let the script crash normally or if you used Write-Host statements instead.

### Reproduction Steps

Put this script into a windows fleet host config:
```
Write-Host ' Start of Host Config Script '
Write-Host ' I am about to intentionally throw an error '

this line isn't a valid powershell command. It's just a regular sentence.

Write-Host ' This line should never be run since it comes after the error '
```

### Environment

Windows SMF + the After Effects + Red Giant persistent storage host config script.

yuanmich2:
Merged the base and persistent-volume scripts into one, improved logging and error handling, and restructured the code so that it leaves less room for error and is more readable

Fixes: https://github.com/aws-deadline/deadline-cloud-samples/issues/243

The old scripts were fragile. The expected filenames was not documented and when it was off the whole script would crash with a generic error. Also the scripts were redundant since the persistent version of the script also handled the case when persistent volumes is disabled.

I merged both scripts. I used a trap block to capture and log errors. I reduced the number of variables that needed to be updated and preferred discovering variable values over hardcoding.

The script should be much more robust and easier to use

I created an After Effects scene with Red Giant and rendered it. I was not able to test the other plugins but their install logic remains largely the same.

Here's the output: 

https://github.com/user-attachments/assets/c5df8182-5c69-4e52-b636-d8b3d4438b6b

I also updated the README file.

---


crowecawcaw (approved):

host_configuration_scripts/aftereffects/aftereffects_redgiant/install-software.ps1
```diff
@@ -1,176 +1,289 @@
...
-    if ($RSMB_HAS_LICENSE) {
-        $RSMB_LICENSING = "FloatingLicensing.zip"
-        # Custom licensing required
-        $RSMB_LICENSE_SERVER = ""
+# CONFIG ================
+# Required
+$AE_INSTALLER_S3_URI = "s3://<bucket>/After Effects_en_US_WIN_64.zip"  # required
+
+# Optional components. Leave blank to skip.
+$RED_GIANT_S3_URI = ""  # e.g. s3://<bucket>/RedGiant-2026.3.0-Win.exe
```
viknith, line 9: Why is RG optional if this is AE + RG script?

host_configuration_scripts/aftereffects/aftereffects_redgiant/install-software.ps1
```diff
@@ -1,176 +1,289 @@
-# Sequential Software Installation Script
-# Downloads installers from S3 and installs Adobe After Effects, Red Giant, Boris Sapphire (optional),
-# Lenscare (optional), and RSMB (optional) in order
-Set-PSDebug -Trace 2
 $ErrorActionPreference = "Stop"
+trap { Write-Output "ERROR: $($_.Exception.Message)`n$($_.InvocationInfo.PositionMessage)`n$($_.ScriptStackTrace)"; exit 1 }
```
viknith, line 2: Did this trap help with error logging? Got an example?

host_configuration_scripts/aftereffects/aftereffects_redgiant/install-software.ps1
yuanmich2: This script includes a lot of code for other plugins too. I left Red Giant as optional just in case someone only needs AE + Boris Sapphire or AE + Real Smart Motion Blur or any combination of plugins. Maybe we should rename it to AE + plugins?

host_configuration_scripts/aftereffects/aftereffects_redgiant/install-software.ps1
yuanmich2: Yes. Before you just get a "Powershell exited with 1" message that made it very difficult to diagnose what went wrong:
image

But the new logs gives more details and shows a trace so you know which lines caused the error. This is the log of the exact same error as the previous screenshot:
image

PR #255, opened 2026-07-14
ci: add static validation checks for sample assets

crowecawcaw:
## What

Adds an offline `pytest` suite (`tests/`) plus a GitHub Actions workflow (`.github/workflows/static_checks.yml`) that runs it on every PR, with pip caching.

## Why

A fleet host configuration script that exceeds the service's **15,000-character** `scriptBody` limit looks fine in the repo but is rejected the moment a customer applies it via `UpdateFleet` — a failure that has previously reached a customer. These static checks catch that class of problem in CI. (For calibration: `queue_environments/conda_queue_env_improved_caching.yaml` is already 14,590 bytes against the 15,000 `EnvironmentTemplate` limit, and `host_configuration_scripts/aftereffects/.../install-software.ps1` is 13,226 bytes.)

## Checks

- **OpenJD templates** — every job & environment template passes `openjd check`.
- **Host configuration scripts** — within the `HostConfiguration.scriptBody` size limit (15,000); Linux `*.sh` pass `bash -n`.
- **Queue environments** — within the `EnvironmentTemplate` size limit (15,000).
- **CloudFormation** — parses (incl. `!Sub`/`!Ref` intrinsics) and passes `cfn-lint` (errors only).
- **Conda recipes** — well-formed `deadline-cloud.yaml` + parseable recipe.

Service limits are pinned to the AWS Deadline Cloud API model in `tests/service_limits.py`.

## Also included

- `fix:` `conda_recipes/conda_build_linux_package/template.yaml` used `{{Param.RecipeDir}}` in the top-level `name` field where a PATH param isn't resolvable; `openjd check` (now enforced) flags it. Changed to `{{RawParam.RecipeDir}}`.

## Testing

`python -m pytest tests -v` passes locally (243 + 8 cfn-lint) and in CI on my fork (239 passed). All fork checks green: Static validation, CodeQL, zizmor, collect.

🤖 Draft — not yet published for review.


tests/test_conda_recipes.py
```diff
@@ -0,0 +1,281 @@
...
+        "os": __import__("os"),
+    }
+
+    # autoescape is enabled to satisfy static analysis; it only affects the
+    # output of ``{{ }}`` expressions (never the literal template text). Every
+    # substitution here is an inert stand-in or a version/name string with no
+    # HTML-special characters, so escaping does not change the parsed YAML.
+    env = jinja2.Environment(
+        undefined=_Undefined, keep_trailing_newline=True, autoescape=True
+    )
```
andychoquette, line 261: i had my agent look into arbitrary code execution in this environment, and it looks like a `recipe/meta.yaml` containing something like `{{ os.popen('...').read() }}` can execute arbitrary code on the runner. We seem safe here since we are careful about what is consumed in this environment (read only token, no secrets) and who can trigger the runner, but there's still a chance of runner abuse. 

The agent's recommended fix: 

Fix: use jinja2.sandbox.SandboxedEnvironment and replace the real os with an inert stand-in (a small dict exposing just os.environ/os.sep as strings, matching the other inert stand-ins already there). That preserves the structural validation while removing the exec path.

crowecawcaw: Technically true, but all the code that runs in the CI check is arbitrary untrusted code. The new check even runs on this PR before the code is merged. Normal PR actions run in sandboxes with no permissions of any kind. It's possible to run PR actions with secrets which does need closer attention, but it's not what we're doing here and GitHub makes it intentionally difficult to set up.

andychoquette (approved):

Issue #265, opened 2026-07-28
Feature request: example CDK stack for the starter farm


crowecawcaw:
### Describe the problem

It'd be nice to have a CDK example for the starter farm. Should be built as a CI action for validation.

### Proposed Solution

...

### Example Use Cases

...

PR #268, opened 2026-07-29
feat(cdk): add AWS CDK starter farm samples

crowecawcaw:
Fixes: *https://github.com/aws-deadline/deadline-cloud-samples/issues/265*

No CDK farm sample existed alongside the CloudFormation and Terraform ones.

Reusable constructs in `lib/deadline/` (`Farm`, `Queue`, `ServiceManagedFleet`, `JobAttachmentsBucket`, `CondaChannel`, `CondaQueueEnvironment`), plus **four example farms** that compose them. They're alternatives — deploy the closest one and copy its source.

| Stack | What you get |
|---|---|
| **`SimpleFarm`** | One queue, one Linux fleet. **Start here** — jobs install apps immediately, nothing to set up first |
| `StarterFarm` | Adds a private Conda channel + a queue that builds packages for it |
| `CudaFarm` | A GPU fleet with `conda-forge` |
| `MultiPlatformFarm` | One queue over Linux, Windows, and GPU fleets |

A whole farm is four constructs; each one owns the IAM its resource needs, and exposes its L1 escape hatch:

```ts
const farm = new Farm(this, 'Farm', { displayName: 'My Farm' });
const bucket = new JobAttachmentsBucket(this, 'Bucket');
const queue = new Queue(this, 'Queue', { farm, displayName: 'Job Queue', jobAttachmentsBucket: bucket });
queue.associateFleet(new CpuLinuxFleet(this, 'Fleet', { farm }));
```

Additive: a new `cdk/` tree and CI workflow. Two notes for reviewers:

- **This app creates the job attachments bucket** (the other samples take an existing one), and deletes it with the stack so a trial farm leaves no storage behind. A farm holding real work passes `RemovalPolicy.RETAIN` to its `JobAttachmentsBucket`.
- **Deadline Cloud allows 3 farms per account.** Fine in normal use since these are alternatives, but I hit it while testing all four.

**All four deployed to a test account with real jobs run on each**, then torn down:

| Farm | Result |
|---|---|
| `SimpleFarm` | Job SUCCEEDED, output returned via job attachments |
| `StarterFarm` | Built `autodock-vina` → published to the private channel → installed and ran it → **`AutoDock Vina v1.2.5`** |
| `CudaFarm` | Landed on an **NVIDIA L4** (CUDA 13.0), matching the preset's `a10g`/`l4` selection |
| `MultiPlatformFarm` | 3-step job: Windows and GPU steps each routed to the only fleet that could satisfy them |

The `StarterFarm` run is the meaningful one — it proves the private-channel read grant works and that the package build queue's parameter separation is right.

**Static:** `tsc` clean, 71 jest tests, `cfn-lint` clean over all four templates, 320 pytest, Vale clean. CI discovers apps via `cdk.json` and a bare `cdk synth` renders every stack, so new stacks need no workflow change. No AWS credentials used.

<details>
<summary>Details on design decisions and known limits</summary>

- **A queue comes with a Conda queue environment** (`addDefaultCondaQueueEnvironment`, default `true`), and passing a `CondaChannel` also grants read access — forgetting that fails at Conda install time, not synth.
- **Each fleet gets its own worker role**, so a grant to one fleet can't reach another's workers.
- **`AcceleratorRequirement` is hand-rolled** rather than reusing the L1 `AcceleratorCapabilitiesProperty`, which lets you give accelerators differing runtimes that the service rejects. The hardware ranges *are* L1 aliases.
- **The package build queue opts out of the Conda environment.** The `conda_recipes/` bundles carry their own and define a `CondaChannels` parameter, so one there would mean two things by one name. Matches the CloudFormation/Terraform farms.
- **Conda grants are scoped per channel and their Sid encodes the channel location**, since IAM needs unique Sids and one queue can hold several channels.
- **No CDK context flags.** Each stack's configuration lives in code in `bin/starter-farm.ts` and its stack file, so there is one place to read and to change.
- **`s3:ListBucket` is bucket-wide**, as in the CloudFormation and Terraform farms — S3 authorizes listing against the bucket, so key names outside a queue's prefix stay listable. Object reads/writes are prefix-scoped. Tightening it with an `s3:prefix` condition may be worth doing, but needs testing against job attachments first.
- **`MultiPlatformFarm`'s `LinuxStep` only requires `os.family: linux`**, which both Linux fleets satisfy, so the scheduler may place it on either. Documented in the README; a property of `hostRequirements`, not the constructs.
- The queue environment stays a standalone OpenJD YAML file byte-identical to the `queue_environments/` copy, so existing `openjd check` discovery validates it. `CondaQueueEnvironment` renders it at synth and bounds it against the 15000-char service limit.

</details>

Yes — `cdk/farm_templates/starter_farm/README.md` (the four farms, composing constructs, deployment, customization), `cdk/README.md` (index + tooling comparison), and TSDoc on every construct explaining *why*.

---


cdk/farm_templates/starter_farm/lib/deadline/queue-environment.ts
```diff
@@ -0,0 +1,170 @@
...
+      `The Conda queue environment template does not contain ` +
+        `${JSON.stringify(CHANNELS_PLACEHOLDER)}, so its default CondaChannels value ` +
+        'cannot be set. Has the queue environment template changed?',
+    );
+  }
+
+  const urls = channels.map((channel) =>
+    typeof channel === 'string' ? channel : channel.url,
+  );
+  return template.replace(
```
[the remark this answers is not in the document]
crowecawcaw: Good catch, and confirmed: with a string replacement, a channel named `foo$&bar` renders as `s3://b/Conda/foodefault: "deadline-cloud"bar deadline-cloud`, splicing the matched placeholder into the URL.

Fixed in 12efe43 with the function replacer you suggested, plus a regression test (`a channel name containing $ is substituted verbatim`) that asserts the literal `$&` survives. The two other `.replace()` calls in `lib/` both replace with `''`, so they cannot hit this.

This comment landed against ea6c8fc, a few minutes before that push.

conda_recipes/cinema4d-openjd/recipe/recipe.yaml
```diff
@@ -1,11 +1,12 @@
...
-# in the target channel. The standalone deadline sample may track a newer,
-# incompatible minor release.
+# Before building this recipe, build the conda
+# packages for "deadline" and "openjd-adaptor-runtime" dependencies.
+# Find these recipes for building the packages at :
+# https://github.com/aws-deadline/deadline-cloud-samples/tree/mainline/conda_recipes
 
 context:
-  version: "0.11.4"
+  version: "0.10.0"
```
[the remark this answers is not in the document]
crowecawcaw: This one is a false positive: this branch does not touch `conda_recipes/` at all.

```
$ git diff --stat upstream/mainline...HEAD -- conda_recipes/
(no output)
```

The changed paths are limited to `cdk/`, `.github/workflows/cdk_checks.yml`, `tests/`, and a few READMEs. The recipe is unchanged at whatever `mainline` holds, so there is nothing to drop here. Looks like the review diffed against a stale base.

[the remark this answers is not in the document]
crowecawcaw: Correction to my reply above — I had the diagnosis half right. The recipe changes were not authored here, but the PR *would* have reverted them if merged.

Root cause: this branch was cut before #267 ("Update Cinema 4D recipes for plugin sync") landed on mainline, so it still carried the pre-#267 recipe at `0.10.0`. GitHub had merged mainline into the PR head (57c158d), and the review compared that against current mainline, so the older file read as a deliberate downgrade. My earlier `git diff` against a stale local `upstream/mainline` is what made it look clean on my end.

Fixed by rebasing onto 6603777 (8d543d3). The recipe is back at `0.11.4` with its `entry_points` block, and the PR's changed-file list is 32 files with zero under `conda_recipes/`. Good catch — this would have silently reverted #267.

cdk/farm_templates/starter_farm/lib/deadline/fleet.ts
```diff
@@ -0,0 +1,381 @@
...
+  readonly cpuArchitecture?: 'x86_64' | 'arm64';
+  /**
+   * Whether to run on spot or on-demand instances.
+   *
+   * Spot instances cost less but can be interrupted, which Deadline Cloud
+   * handles by retrying the interrupted task on another worker.
+   *
+   * @default 'spot'
+   */
+  readonly instanceMarketType?: 'on-demand' | 'spot';
```
karthikbekalp: And wait and save?

cdk/farm_templates/starter_farm/lib/deadline/fleet.ts
```diff
@@ -0,0 +1,381 @@
...
+  constructor(scope: Construct, id: string, props: PresetFleetProps) {
+    super(scope, id, {
+      ...preset(props, {
+        displayName: 'CUDA Linux Fleet',
+        osFamily: 'LINUX',
+        instanceMarketType: 'on-demand',
+        maxWorkerCount: 1,
+        vCpuCount: { min: 4, max: 16 },
+        memoryMiB: { min: 32768 },
+        accelerators: { names: ['a10g', 'l4'] },
```
karthikbekalp, line 366: Why only these and not T4 or L4S? 

I think T4s are cheaper but very not that efficient. L4S are expensive though.

karthikbekalp (commented):
Just a bunch of nits on the options available.

cdk/farm_templates/starter_farm/lib/deadline/fleet.ts
crowecawcaw: Exactly that. I mirrored the starter farm in our CFN and Terraform examples, and I think that's the reason they ignored T4 and L4S.

karthikbekalp: Cool.

karthikbekalp (dismissed):

karthikbekalp (approved):

PR #270, opened 2026-07-30
feat: add Rez PATH-shim queue environment with tool auto-discovery

crowecawcaw:
Fixes: *no issue yet — opening as a draft per CONTRIBUTING's "talk with us first" guidance, happy to file one if you'd prefer to track it there*

A customer reported that Rez does not work well with queue environments, because a queue environment runs in a single subprocess and the render command only receives environment variables harvested from it and replayed.

That is accurate, and it is structural rather than a bug. A queue environment action runs in its own subprocess, so the only channel into later actions is `openjd_env: NAME=value` directives. The existing [`rez_queue_env.yaml`](https://github.com/aws-deadline/deadline-cloud-samples/blob/mainline/queue_environments/rez_queue_env.yaml) therefore activates a context, diffs the environment before and after, and replays the difference. Anything that is not a name-value pair does not survive.

A Rez `alias` is the clearest casualty. Rez implements it as an exported shell function, which Bash exports under a name like `BASH_FUNC_launch%%` with a multi-line value. The session runtime rejects that assignment outright:

```
openjd_env: "BASH_FUNC_demoalias%%=() {  demorender --via-alias \"$@\"\n}"
  -- ERROR: Failed to parse environment variable assignment.
```

So the alias is gone before any task runs. Ordered `PATH` edits are similarly fragile: the variable is restored, but a later environment that also edits `PATH` can reorder it.

A queue environment that wraps each task's command instead of exporting variables to it.

`onEnter` resolves the requested packages once, saves the context to a `.rxt` file in the session directory, asks Rez which executables those packages provide, and writes one shim per tool with the shim directory prepended to `PATH`:

```bash
#!/usr/bin/env bash
exec rez env --input "$REZ_CONTEXT_FILE" --shell bash -- "<tool>" "$@"
```

Job templates keep calling tools by bare name, such as `command: mayapy`, so each call re-enters the saved context in its own shell and Rez applies the full context inside the task's own process. **Job bundles need no changes.**

Tool names come from `rez context -t` on the saved context, so there is no hard-coded list of DCC executables. `RezExtraTools` covers commands a package provides without declaring them in its `tools` list.

This is a workaround for a specification gap, not a permanent design. [RFC0008: Environment Wrap Actions](https://github.com/OpenJobDescription/openjd-specifications/issues/132) proposes `onWrapTaskRun`, which does this properly; once it ships in the worker agent, the wrap hook replaces the shim directory and the `PATH` manipulation entirely. The README says so explicitly so nobody mistakes this for the intended long-term shape.

Additive. Three new files under `queue_environments/rez_shim/`, no changes to existing samples:

| File | Purpose |
|---|---|
| `rez_queue_env_shim.yaml` | The queue environment to deploy on a farm |
| `rez_demo_setup_queue_env.yaml` | Test scaffolding: installs Rez and builds a demo package |
| `demo_job_bundle/template.yaml` | Verification job |

The existing `rez_queue_env.yaml` is untouched and remains the right choice when packages only set variables, and the only choice on Windows.

Known tradeoffs, all documented in the README:

- Only bare command names are intercepted; an absolute path in a template bypasses the shims.
- **Linux and macOS workers only.** The shims are POSIX scripts relying on a shebang, so the environment fails immediately on Windows with a message pointing at `rez_queue_env.yaml`. I deliberately did not claim Windows support, since it is not tested.
- Each task pays a context re-entry, mitigated by Rez's resolve cache.
- The demo's package repository is created outside the session directory and persists on the worker; documented under cleanup.

Locally with `openjd run`, and end to end on a Deadline Cloud farm with a Linux service-managed fleet (on-demand, 2–4 vCPU). All templates pass `openjd check`, and `python3 scripts/validate_repository.py` passes.

The demo job proves its own thesis rather than asserting it. `VerifyEnvironment` runs three fidelity checks and fails the task on regression:

| Check | State under test | Under harvest-and-replay |
|---|---|---|
| 1 | A plain variable | Survives |
| 2 | A Rez `alias` | Lost, rejected by the runtime |
| 3 | A `PATH` prepend shipping its own `sort` | Depends on environment order |

Farm run, on the worker under `job-user`:

```
--------- Entering Environment: RezDemoSetup
Created demo Rez package 'demotool-1.0.0' in /tmp/rez-demo-repository
--------- Entering Environment: Rez
Created 3 Rez tool shims in /sessions/session-.../rez-shims:
   democonvert
   demorender
   demosleep
=== DEMOTOOL_VERSION as seen by the session (expected UNSET) ===
DEMOTOOL_VERSION=UNSET
=== DEMOTOOL_VERSION as seen inside the tool (expected 1.0.0) ===
demorender: DEMOTOOL_VERSION=1.0.0
=== Check 1: plain variable reaches the tool ===
PASS: variable visible inside the tool
=== Check 2: Rez alias survives into the task ===
PASS: alias is callable
=== Check 3: package PATH prepend shadows the system command ===
PASS: package command shadows the system one
All 3 environment fidelity checks passed.
```

`UNSET` in the session with `1.0.0` inside the tool is the core result: the variable was never harvested and replayed, Rez applied it inside the task's own process.

**Cancelation was verified too**, since it is the obvious risk with a wrapper. Rez runs the tool in a shell of its own, so the tree is `shim` → `rez env` → shell → tool rather than flat, but signals still propagate. Canceling the `CancelThroughShim` step on the farm:

```
demosleep: pid=39289 sleeping for 600s
Canceling subprocess 39247 via notify then terminate method
INTERRUPT: Sending signal "term" to process 39247
demosleep: caught SIGTERM, exiting
Process pid 39247 exited with code: 1
```

Task ended `CANCELED` with no orphans, so an application with its own signal handler gets to shut down cleanly.

Not tested: Windows (unsupported by design, guarded with a clear failure), and macOS workers on a real fleet — the macOS path was exercised locally only.

Yes. One README for the whole set at `queue_environments/rez_shim/README.md`, covering when to choose this over `rez_queue_env.yaml`, why the simpler sample cannot cover those cases, deployment, parameters, tradeoffs, the upstream RFC, and cleanup. The `queue_environments` index table points at the directory, matching how nested collections are handled elsewhere in the repo.

---


crowecawcaw:
Good catch, and fixed in 5c6f003 — though the mechanism is narrower than described, so I want to record what I found.

**The two scenarios in the comment do not actually recurse.** Rez rebuilds `PATH` from the resolved context rather than inheriting it, so `$SHIM_DIR` is not on `PATH` inside `rez env` at all. I tested both cases against real Rez packages:

- `env.PATH.append("{root}/bin")` — inside the context, `PATH` starts at the package's `bin` and contains no shim directory. `command -v appdo` → `/private/tmp/recurse/repo/apptool/1.0.0/bin/appdo`. The shim ran once and exited cleanly.
- A package declaring `tools = ["sort"]` without providing it — resolves to `/usr/bin/sort`. A `RezExtraTools` name that exists nowhere fails with a plain `command not found`, not a loop.

**But the underlying risk is real under a different trigger.** If a farm's Rez config lists `PATH` in [`parent_variables`](https://rez.readthedocs.io/en/stable/configuring_rez.html), Rez merges the parent `PATH` instead of replacing it, `$SHIM_DIR` stays first, and the shim does find itself. Reproduced with `REZ_PARENT_VARIABLES=PATH`: **60 processes in 15 seconds** before I killed it. Not the default, but plausible in a studio that expects farm tooling to remain visible inside a context — and the failure mode is as bad as described.

So the suggested remedy was right even though the diagnosis was not. Each tool is now resolved once with `command -v` inside the context when its shim is written, and the shim `exec`s that absolute path:

```bash
exec rez env --input "$CTX" --shell bash -- "/abs/path/to/tool" "$@"
```

Anything resolving back into `$SHIM_DIR`, or unresolvable in the context, gets no shim and is reported at startup rather than failing at task time. Verified that the guard refuses exactly the `parent_variables` case (`GUARD: refused`) while accepting the real binary under a default config.

Re-verified end to end on a Linux service-managed fleet after the change: 3 shims created, no warnings, all 4 tasks succeeded, all 3 fidelity checks passed.

crowecawcaw:
Confirmed and fixed in 691a632. This one was right on the substance, and I could reproduce the weakness directly.

**The falsifying test.** I removed the `openjd_env: PATH=$SHIM_DIR:$PATH` line from the environment, so no shim was ever on `PATH`, and re-ran the verify step against the old checks:

```
SABOTAGE: deliberately not injecting SHIM_DIR into PATH
=== Check 1 === FAIL: variable missing inside the tool
=== Check 2 === PASS: alias is callable
=== Check 3 === PASS: package command shadows the system one
openjd_fail: 1 of 3 environment fidelity checks failed.
```

Exactly as described: with the mechanism completely broken, two of three checks still passed. They were testing that the `.rxt` file could be re-entered, which is a property of Rez, not of this sample.

**The fix.** The demo package now provides a `demoprobe` tool that reports the alias and `sort` resolution from inside a correctly applied context. Because it is listed in the package's `tools`, it gets a shim like any other tool, so the verify step calls it by bare name and reads both results from its output. No check invokes `rez env --input` any more.

Same sabotage against the new checks:

```
=== Check 1 === FAIL: variable missing inside the tool
=== Check 2 === FAIL: alias was lost
=== Check 3 === FAIL: system command won, PATH order was not preserved
openjd_fail: 3 of 3 environment fidelity checks failed.
```

All three now depend on shim delivery end to end, so "All 3 environment fidelity checks passed" means what it says. I added a note in `verify.sh` explaining why the probe is indirect, and a line in the README recording that deleting the `PATH` injection fails all three — that being how the checks were validated.

Re-verified unbroken on a Linux service-managed fleet: 4 shims created, all 4 tasks succeeded, all 3 checks passed.

crowecawcaw:
Correct, fixed in ae9b45c. The comment was carried over from `rez_queue_env.yaml` and contradicted the environment's own behavior: it told you to set a Windows path, while the script defines only `LINUX_REZ_REPOSITORY_PATH` and `MACOS_REZ_REPOSITORY_PATH` and fails on non-Linux/Darwin workers. Now reads "the Linux and MacOS repository path".

I checked the other two Windows mentions in the file and left them alone — both are the intentional guard (the shebang explanation and the `openjd_fail` message pointing at `rez_queue_env.yaml`).

leon-li-inspire (approved):

PR #275, opened 2026-08-06
feat: add germline variant calling job bundle using bioconda

mwiebe:

The repository had no sequence-bioinformatics sample. The existing life-sciences
samples cover protein structure prediction ([`esmfold_predict`](job_bundles/esmfold_predict/)),
molecular dynamics ([`gromacs_md`](job_bundles/gromacs_md/)), and molecular docking
([`virtual_screening_vina/`](job_bundles/virtual_screening_vina/)), but nothing that
starts from DNA sequencing reads. Variant calling is the workhorse analysis in genomics,
and it suits Deadline Cloud well because both expensive stages divide cleanly: samples
align independently, and separate regions of the genome are called independently.

No sample sourced software from the [bioconda](https://bioconda.github.io/) channel
either. Nearly all open source bioinformatics tooling is published there, so a customer
evaluating Deadline Cloud for genomics had no worked example of the channel setup that
this domain requires.

Adds `job_bundles/variant_calling_bwa/`, a germline variant calling pipeline that goes
from paired-end FASTQ reads to a merged VCF using `bwa`, `samtools`, and `bcftools`.

Five steps across two independent axes of parallelism:

```
QualityControl  (1 task/sample)     BuildIndex  (1 task)
  fastqc on each read pair            samtools faidx + bwa index
           |                                  |
           |                        AlignReads  (1 task/sample)
           |                          bwa mem | samtools sort
           |                                  |
           |                        CallVariants  (1 task/region)  <- the scatter
           |                          bcftools mpileup | call | filter
           +----------------+-----------------+
                            |
                  MergeVariants  (1 task)                          <- the gather
                    bcftools concat | norm, stats, MultiQC
```

`AlignReads` fans out over samples and `CallVariants` over genome regions, using
different task parameters in different steps. With the shipped defaults that comes to
10 tasks.

The tool sequence follows the
[AWS HealthOmics WDL variant-calling tutorial pipeline](https://github.com/aws-samples/aws-healthomics-tutorials/tree/main/example-workflows/wdl/variant-calling-pipeline),
reimplemented for Open Job Description. Having the same pipeline expressed in both
specifications makes their designs comparable. The WDL version runs each tool in its own
BioContainers image; because those images are themselves built from bioconda packages,
the tool builds are equivalent while the delivery mechanism collapses to a single
`CondaPackages` line. This bundle also adds the second scatter over regions, which the
original lacks, and takes paired-end reads where the original takes one FASTQ per sample.

Design decisions worth calling out, each recorded in the commit message and the sample
README:

* **`CondaChannels` defaults to `conda-forge bioconda`, in that order.** bioconda depends
  on conda-forge for its runtime libraries and documents that priority. Reversing the
  order causes resolution failures that are hard to read.
* **The reference is a `ReferenceDir` directory plus a `ReferenceFastaName` filename**
  rather than one `FILE`-typed parameter. Job attachments stages exactly the path a
  `FILE` parameter names and does not sweep in siblings, so a `.fai` or bwa index built
  beside the FASTA reaches the worker only if the whole directory is staged.
* **The shell lives in `scripts/`** rather than inline in the template, following stage 2
  of the [job development progression](job_bundles/job_dev_progression/). At roughly 400
  lines the bundle is past the point where a self-contained template stays readable.
  `scripts/common.sh` holds the per-region VCF filename derivation that `CallVariants`
  and `MergeVariants` must agree on.
* **Every path is derived from a job parameter** rather than passed between steps, because
  each step runs in its own session directory, and **each step declares every step whose
  output it reads**, because dependency entitlements do not chain.

Test data is downloaded by `sample_inputs/fetch_test_data.py` (about 950 KB from the
public `aws-genomics-static-us-east-1` bucket, originating from nf-core/test-datasets,
MIT licensed) rather than committed.

The change also fixes an unrelated repository validation failure: `check_external_links.py`
now reports link locations with `as_posix()`, because `str()` on a relative path yields
`docs\guide.md` on a Windows checkout and the test expected forward slashes.

Additive. One new job bundle, one row in the `job_bundles/README.md` category table, and
a one-line fix in `scripts/check_external_links.py`. No existing sample changes behavior.

The bundle requires a Linux fleet and a conda queue environment whose channels include
conda-forge and bioconda. When a tool is missing, the `BioToolchain` job environment
fails on entry and names the exact packages and channels to configure, rather than
failing partway into a step.

Running the job incurs Deadline Cloud worker, storage, and data transfer charges. The
README documents that, and notes that genomic sequence data is often subject to consent,
privacy, and jurisdictional restrictions, so access controls deserve attention before the
bundle is pointed at real data.

**Submitted to a Deadline Cloud farm with the shipped defaults.** All 10 tasks succeeded
on a Linux service-managed fleet in about 100 seconds of wall clock. The merged VCF held
variants for both samples:

```
=== Variant Calling Summary ===
Regions called: 4
Total variants: 17
Samples:        tiny_n tiny_t

--- variants per contig ---
  1            17

--- bcftools stats ---
number of samples:	2
number of records:	17
number of SNPs:	17
number of indels:	0
```

The alignment is genuinely good rather than vacuous. `samtools flagstat` for `tiny_n`:

```
1174 + 0 in total (QC-passed reads + QC-failed reads)
1164 + 0 mapped (99.15% : N/A)
1144 + 0 properly paired (97.44% : N/A)
```

Every expected artifact landed: `variants.vcf.gz` and its `.tbi`, four per-region VCFs
each with an index, two sorted BAMs with `.bai` and flagstat summaries, four FastQC
reports, and a MultiQC report. MultiQC's source manifest confirms it parsed the real
FastQC archives and flagstat files.

The bioconda path is confirmed in the worker log, which shows the job's channel override
taking effect over the queue environment's default:

```
CondaChannels: conda-forge bioconda
CondaPackages: bwa samtools bcftools fastqc multiqc
```

**Repository validation.** `python3 scripts/validate_repository.py` passes (35 unit tests,
static local-link checks). `python3 scripts/check_external_links.py` passes, which the
`as_posix()` fix was needed for on a Windows checkout.

**Template and script checks.** The template decodes and builds a job through the
`openjd` model library, producing the documented task counts (`BuildIndex` 1,
`QualityControl` 2, `AlignReads` 2, `CallVariants` 4, `MergeVariants` 1). `bash -n` is
clean on all seven scripts.

**Shared helper behavior**, exercised directly under `set -euo pipefail`, the mode the
scripts run in:

| Input to `parse_list` | Result |
|---|---|
| `"tiny_n, tiny_t"` | `[tiny_n]` `[tiny_t]` |
| `"  lead ,  both  , trail  "` | `[lead]` `[both]` `[trail]` |
| `"1:131000-133999, 1:134000-136999"` | region strings preserved intact |

Trimming only the surrounding whitespace matters: deleting interior spaces too would
rewrite a name like `tiny n` to `tinyn` before `validate_sample_names` saw it, so the job
would fail on a filename the user never typed. An empty list parameter is rejected up
front, because zero entries would otherwise be reported by every per-index task as past
the end of the list while the gather found nothing missing, and the failure would surface
inside `bcftools` handed no BAM arguments.

**Not covered by this testing:** the run reused a cached named conda environment, so it
did not exercise a cold bioconda solve. An earlier end-to-end run did cover that path.

Yes.

* [`job_bundles/variant_calling_bwa/README.md`](job_bundles/variant_calling_bwa/README.md)
  covers what the sample demonstrates, prerequisites, the conda channel order and why it
  matters, setup, local iteration with the Open Job Description CLI, a full parameter and
  output table, how steps pass files to each other, why the `...Range` parameters exist,
  security and cost, and troubleshooting. It also states what the sample leaves out that
  a production analysis would add (duplicate marking, base quality score recalibration,
  adapter trimming, per-contig ploidy), so nobody mistakes it for a clinical-grade
  pipeline.
* [`job_bundles/variant_calling_bwa/sample_inputs/README.md`](job_bundles/variant_calling_bwa/sample_inputs/README.md)
  documents the test data provenance and licensing.
* The new row in [`job_bundles/README.md`](job_bundles/README.md) follows the existing
  category table format.
* Root navigation is unchanged, because no recommended starting point changed.

---

### DIFF
--- a/job_bundles/README.md
+++ b/job_bundles/README.md
@@ -60,6 +60,7 @@ Nested collections provide their own complete indexes where applicable.
 | [Maya V-Ray tiled render](tile_render_with_maya_vray/) | V-Ray tile rendering followed by OpenImageIO assembly | You need tiled EXR output from Maya and V-Ray |
 | [V-Ray Linux region render](tile_render_with_vray_linux/) | Region rendering, asset discovery, path mapping, and image merge | You render `.vrscene` files in parallel regions on Linux |
 | [Maya Arnold turntable](turntable_with_maya_arnold/) | Building a scene around an OBJ, then rendering frames and encoding video | You need an easy-to-submit 3D asset review utility |
+| [Germline variant calling](variant_calling_bwa/) | DNA sequence alignment and variant calling with bioconda software, scattered per sample and per genome region | You want to find genetic differences from sequencing data, or a sequence-bioinformatics starting point |
 | [Studio VFX shot pipeline](vfx_pipeline/) | A launcher, static render bundle, Conda software delivery, and Flow publishing mapped onto a studio's existing pipeline | You are porting a studio shot pipeline to Deadline Cloud |
 | [AutoDock Vina virtual screening](virtual_screening_vina/) | Parallel molecular docking and ranked result aggregation | You want to screen many ligands against a protein target |
 | [vLLM evaluation leaderboard](vllm_lm_eval_leaderboard/) | Parallel model evaluation and final CSV/Markdown aggregation | You need to compare multiple LLMs across benchmarks |
--- a/job_bundles/variant_calling_bwa/README.md
+++ b/job_bundles/variant_calling_bwa/README.md
@@ -1,0 +1,494 @@
+# Germline variant calling with bwa, samtools, and bcftools
+
+Finds the genetic differences between an individual and a reference genome.
+
+A DNA sequencing machine does not read a genome end to end. It produces hundreds of millions of
+short fragments, each a few hundred letters long, from random positions. Turning those fragments
+into a list of differences takes three stages, and this job bundle runs all three:
+
+1. **Align.** Work out where each fragment came from by matching it against a reference genome.
+   `bwa` does this.
+2. **Call.** At every position, compare the stacked-up fragments to the reference and decide whether
+   the difference is real or a sequencing error. `bcftools` does this.
+3. **Merge.** Collect the results into one file listing every confident difference.
+
+The differences are **variants**, and finding them is **variant calling**, the workhorse analysis
+behind everything from diagnosing inherited disease to breeding drought-tolerant crops. *Germline*
+means the variants an individual was born with, rather than ones a tumor acquired later. The input
+format is **FASTQ** (the sequencer's fragments) and the output a **VCF** (the list of variants).
+
+Both expensive stages divide cleanly, which is what suits this to Deadline Cloud. Each sample aligns
+on its own worker, and separate stretches of the genome are then called independently and stitched
+back together.
+
+## What this sample demonstrates
+
+* **Software from bioconda.** `bwa`, `samtools`, `bcftools`, and `fastqc` are declared as conda
+  packages rather than built into container images. See
+  [Conda channel order](#conda-channel-order) for why `conda-forge` must come first.
+* **Fan-out over samples and over genome regions.** `AlignReads` fans out over samples and
+  `CallVariants` over regions, using different task parameters in different steps. `CallVariants`
+  is the scatter and `MergeVariants` the gather.
+* **A job environment as a precondition check.** `BioToolchain` verifies every command in
+  `RequiredTools` is on `PATH` once per session and, if any is missing, prints the exact packages and
+  channels to configure.
+* **Scripts bundled beside the template.** The shell lives in [`scripts/`](scripts/) rather than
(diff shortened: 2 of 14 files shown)


crowecawcaw (approved):
Cool! Does this workflow generate any interesting visuals that we could use as a hero image for the sample README?

mwiebe:
> Cool! Does this workflow generate any interesting visuals that we could use as a hero image for the sample README?

Not that I'm aware of, but I didn't dig into what options the tools it uses have for visualizing results. Would be a nice follow-up!

PR #286, opened 2026-08-27
feat(maya-redshift): support Maya 2027 and add a Redshift render sample

andrew-fenton:

`maya-redshift` could not be used with Maya 2027: it pinned Redshift 2026.2.1 and declared
`maya >=2025, <2027`, so `maya=2027 maya-redshift` failed to solve. Maya 2027 needs Redshift
2026.6.0 or later. There was also no sample for rendering a Maya scene with Redshift.

1. `conda_recipes/maya-redshift-2026`: bump to Redshift 2026.8.1, generate the
   `redshift4maya.mod` module file for Maya 2027, and widen the run requirement to
   `maya >=2025, <2028`.
2. `job_bundles/maya_redshift_render`: new sample rendering with `Render -r redshift`.

The sample scene contains Redshift nodes on purpose. `Render` reads the scene and then runs
Redshift's `melheader`, which needs redshift4maya already loaded; the package installs it as a
Maya module without enabling autoload, so the scene's own `requires "redshift4maya"` statement
is what loads it. A scene without Redshift nodes fails with
`Unknown object type: RedshiftOptions`.

Maya 2027 can use Redshift, and Maya 2025 and 2026 remain supported by the same package.
Rebuilding the recipe now produces Redshift 2026.8.1 instead of 2026.2.1.

Built the package with `submit-package-job maya-redshift-2026 -p linux-64` and submitted the new
bundle on a Linux GPU service-managed fleet, overriding only the conda channel.

| Maya | Result |
|---|---|
| 2027 | `[Redshift] Redshift for Maya 2027`, `License acquired`, EXR written, 5.25s |
| 2025 | `[Redshift] Redshift for Maya 2025`, `License acquired`, EXR written, 5.55s |

```
[Redshift] Redshift for Maya 2027
[Redshift] License acquired
[Redshift] Saved file '.../output/redshift_spheres.0001.exr' in 0.04s
[Redshift] Frame done - total time for frame 1 (1/1): 5.25s
```

`python3 scripts/validate_repository.py` passes (35 unit tests, 165 Markdown files).

Yes: updated the recipe README (version, Maya 2025-2027, the 2026.6.0 minimum), the
`conda_recipes` and `job_bundles` category tables, and added a README for the new sample.

Happy to split the recipe change and the sample into separate PRs if you prefer; I kept them
together because the sample needs a `maya-redshift` build that supports the Maya version rendered.

---


job_bundles/README.md
```diff
@@ -9,7 +9,7 @@ for more about submitting these jobs to your farm.
 
 ## Job bundle index
 
-This table covers every immediate user-selectable sample directory or collection in `job_bundles/`.
+This table is the complete index of user-selectable sample directories and collections in `job_bundles/`.
```
andrew-fenton, line 12: Had to change this since CI was failing on `Vale` check.

job_bundles/README.md
```diff
@@ -105,4 +106,4 @@ If you do not want to use the `deadline` Python package's support for features l
 ## Example Husk USD render with asset introspection
 
 The [Houdini Husk USD render](houdini_husk_usd_render/) sample shows how to use the Houdini Husk CLI USD renderer using a short job template and service-provided Conda packages.
-It also shows how to write a custom asset introspection tool for job attachments, ensuring that only the required data is uploaded while removing manual steps for artists.
+It also shows how to write a custom tool that introspects assets for job attachments, ensuring that only the required data is uploaded while removing manual steps for artists.
```
andrew-fenton, line 109: Had to change this since CI was failing on `Vale` check.

conda_recipes/maya-redshift-2026/recipe/recipe.yaml
```diff
@@ -2,15 +2,15 @@
 # See https://prefix-dev.github.io/rattler-build/latest/
 
 context:
-  version: "2026.2.1_2202748377"
```
kenyrish-amazon, line 5: I assume we won't need to release a maya-redshift-2027 sample recipe when it comes out (seems like october) since this version does the same thing, right?

job_bundles/maya_redshift_render/template.yaml
```diff
@@ -0,0 +1,148 @@
...
+        set -x
+        # The Redshift renderer accepts a different set of flags than the Maya software
+        # renderer. Run `Render -r redshift -help` to list them.
+        Render -r redshift \
+            -proj '{{Param.ProjectPath}}' \
+            -rd '{{Param.OutputDir}}' \
+            -s {{Task.Param.Frame}} \
+            -e {{Task.Param.Frame}} \
+            -pad 4 \
+            -cam {{Param.CameraName}} \
```
kenyrish-amazon: {{Param.CameraName}} is unquoted here while every other value on this command (-proj, -rd, scene file) is single-quoted. Since it's a user-supplied parameter, a name with a space breaks the render and a value like $(...) runs arbitrary commands on the worker. We should add quotes `-cam '{{Param.CameraName}}' \`

kenyrish-amazon (dismissed):

job_bundles/maya_redshift_render/template.yaml
```diff
@@ -0,0 +1,148 @@
...
+
+# Software Environment
+- name: CondaPackages
+  type: STRING
+  userInterface:
+    # This parameter is for a queue environment, not used directly in the job,
+    # so leave the GUI for the queue environment to display.
+    control: HIDDEN
+  description: Choose which conda packages to install for the render. Requires a conda queue environment to process the value.
+  default: maya maya-redshift
```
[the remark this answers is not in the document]
kenyrish-amazon: We shouldnt have the test scene have `requires -nodeType ... "redshift4maya" "2026.8.1"` if it is meant for all redshift scenes

job_bundles/maya_redshift_render/template.yaml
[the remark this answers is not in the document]
kenyrish-amazon: +1

andrew-fenton: True, will change.

conda_recipes/maya-redshift-2026/recipe/recipe.yaml
andrew-fenton: Correct.

job_bundles/maya_redshift_render/template.yaml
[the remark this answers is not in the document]
andrew-fenton: I tested this, and the version in that statement turns out not to restrict anything - Maya treats it as informational. Maya loads whatever redshift4maya the module path provides; the recorded version only documents what the scene was saved with.

job_bundles/maya_redshift_render/template.yaml
andrew-fenton: The rest of the parameters are integers so we shouldn't need quotes.

kenyrish-amazon (approved):

PR #289, opened 2026-09-09
feat(3dsmax): 3ds Max 2026 + Pencil+ S3-cached host configuration

ZainAallii:
## Summary

Adds a 3ds Max 2026 + Pencil+ 4 host configuration sample for Windows service-managed fleets. The first worker installs once and publishes the result to S3, and later workers restore that payload instead of reinstalling. A fleet persistent volume is an optional accelerator on top of the S3 cache.

New folder `host_configuration_scripts/3dsmax/3dsmax-2026-pencilplus-persistent/`:

- `s3-hosted-3dsmax-2026-pencilplus-persistent.ps1` — the full host configuration script. Boot paths: S3-COLD (install, capture, publish), S3-WARM and PV-WARM (graft), and WAIT (token-guarded concurrency).
- `s3-bootstrap-loader.ps1` — the inline loader, 4 KB, which is what you paste as the fleet's `scriptBody`. It downloads and runs the full script from S3.
- `README.md` — caching model with a Mermaid boot-path diagram, prerequisites, security and trust, configuration reference, setup, and verification.

The parent `host_configuration_scripts/3dsmax/README.md` gains a 2026 index row.

3ds Max 2026 is not affected by the 3ds Max 2027 Autodesk ADP "Failed to start" issue, so no ADP workaround is included.

## Why the script is hosted in S3, and the `tests/` change

[`HostConfiguration.scriptBody`](https://docs.aws.amazon.com/deadline-cloud/latest/APIReference/API_HostConfiguration.html) has a maximum length of 15000 characters. This script is larger than that, so the file you paste into the fleet is the 4 KB loader and the full script is fetched from S3 at boot. The large file is a payload, like the 3ds Max installer zip, and no API limit applies to it.

`test_script_within_service_length_limit` applies the 15000 limit to every `.ps1` under `host_configuration_scripts/`, which assumes anything in that directory becomes `scriptBody`. That holds for every other sample there and not for an S3-hosted one, so the check measures the wrong file.

The change adds coverage rather than relaxing it:

- Scripts named with the `s3-hosted-` prefix are exempt from the byte check. The prefix is in the filename so the exemption is visible in the repository tree rather than in a list.
- The loader is not exempt, so the limit is now enforced on the file that is actually subject to it.
- `test_s3_hosted_script_has_loader` requires a real loader beside the script, identified by content (it references `HC_SCRIPT_S3_URI` and invokes `powershell.exe` on what it downloaded), and requires the script to name it.

Deliberately not a bare `s3` prefix, because the loader is conventionally named `s3-bootstrap-loader.ps1` and it is the one file that must stay within the limit.

The same reasoning already exists one function above, in [`find_utility_scripts()`](https://github.com/aws-deadline/deadline-cloud-samples/blob/a6afff5/tests/conftest.py#L128-L134): scripts that are not uploaded to the service are exempt from the length check while still getting syntax checks.

Happy to change the mechanism if you would rather infer it than name it.

## Testing

Each boot path was exercised on a Windows service-managed fleet:

- S3-COLD: installed 3ds Max 2026 and Pencil+, captured the payload, published the zip to S3.
- S3-WARM: restored the payload on a fresh instance and grafted.
- PV-WARM: grafted from the persistent volume in about 11 seconds.

Host configuration exited 0 on each path, and a Pencil+ scene rendered through `3dsmaxcmd.exe`.

Static checks on this revision: `python3 scripts/validate_repository.py` passes, `vale` reports no findings, and both PowerShell files parse under Windows PowerShell 5.1 and pwsh 7.

## Notes

All S3 URIs use `your-bucket-name` placeholders, and no account, farm, or bucket details are included. The script refuses to run while those placeholders are still in place, rather than failing later with an S3 `AccessDenied` that looks like a permissions problem.

### DIFF
--- a/host_configuration_scripts/3dsmax/3dsmax-2026-pencilplus-persistent/README.md
+++ b/host_configuration_scripts/3dsmax/3dsmax-2026-pencilplus-persistent/README.md
@@ -1,0 +1,184 @@
+# 3ds Max 2026 + Pencil+ 4 host configuration (S3-cached, persistent-volume accelerated)
+
+This host configuration installs 3ds Max 2026 and Pencil+ 4 on AWS Deadline Cloud Windows
+service-managed fleet workers. Unlike the standard scripts, it installs only once. The result is
+captured to S3 and reused on every later worker rather than reinstalled. A fleet persistent
+volume is an optional accelerator on top of the S3 cache.
+
+3ds Max 2026 is not affected by the 3ds Max 2027 Autodesk ADP "Failed to start" issue, so no
+ADP workaround is needed here.
+
+## Files
+
+| File | Purpose |
+|---|---|
+| [`s3-hosted-3dsmax-2026-pencilplus-persistent.ps1`](s3-hosted-3dsmax-2026-pencilplus-persistent.ps1) | The full host configuration script. On a cold boot it installs 3ds Max 2026 + Pencil+ 4 and captures the payload. On warm boots it grafts that payload. |
+| [`s3-bootstrap-loader.ps1`](s3-bootstrap-loader.ps1) | An inline loader that downloads and runs the full script from S3. Use it because the full script exceeds the 15,000-character inline `scriptBody` limit. |
+
+The `s3-hosted-` prefix marks a script that is fetched from S3 at boot rather than pasted into the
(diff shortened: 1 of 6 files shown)


ZainAallii:
Created a new clean PR.

PR #290, opened 2026-09-11
feat: add OpenDroneMap simple job

prestomation:
Add a Python-native, single-task ODM 3.6.1 job bundle for user-selected JPEG surveys, an ODM-specific checksummed run manifest, tests, documentation, and a generic Docker Engine host configuration for CPU container jobs.

Local validation:
- 11 unit tests passed.
- openjd check and openjd summary passed.
- Python compileall, bash syntax, and git diff checks passed.

We want to have a broad set of sample jobs for common industry applications

This is a sample of how to run ODM using docker host configuration

Users who want to process drone imagery at scale have a nice starting point

I ran using a sample data set and received the overhead imagery and report as output

- If the change is a sample, then please describe the steps that you took to test it.

yes
- If applicable, has the sample's description been updated?

---


crowecawcaw (dismissed):
Does this generate a cool image we can add a screenshot of? Is there a standard sample set we can link to in the README for testing?

crowecawcaw (approved):

